Issue #312, opened 2026-04-02
Leverage parallel matching (via meta-matching) to generate longest common subsequences to drive the three-way merges.


sageserpent-open:
> * Going further, the detection of parallel matches actually yields _groups_ of parallel matches. Each group either represents a block of code that stays in place, or moves intra- or inter-file. So if per-file merging is recast as per-group merging for all the groups, then we would have a unified approach to merging that would apply to both unmoved and moved code. Furthermore, this would generalise the notion of splicing, replacing sequences of anchors and splice inputs with a single merge. That might help with the context problem mentioned in the previous point.
> * If we are working with merges per-parallel move group, then the parallel matches in the group can serve as a way of building up an LCS for the subsequent merge (might need to finesse this because meta-matches are distinct for all-sides and pairwise matches - so more than one parallel group might need to be combined to make a mergable group). That would get rid of the need to use the slow classic LCS algorithm.
> 
> Let's expand on how this could be done for posterity.
> 
> A meta-match is translated back into one or more outer (ordinary) matches, and when there are several matches coming from a single meta-match then we have a group of parallel matches. Cool. Where it gets tricky is that meta-matches are distinct for all-sides and pairwise matches, so a mix of outer matches that form an obviously parallel sequence but that are pairwise and all-sides will form corresponding separate meta-matches.
> 
> (This also happens if, say one of several all-sides matches forms part of an breakaway move that is flanked by conformist parallel matches, but that is OK - one expects to see multiple groups in that case).
> 
> Anyway, once the separate all-sides and pairwise meta-matches translate back into groups containing all-sides and pairwise *matches*, we need to fuse these groups together where appropriate.
> 
> Now, the meta-matches aren't subject to reconciliation, so this can be taken advantage of: if a large pairwise meta-match subsumes a small all-sides meta-match, then we expect the back-translated all-sides matches to be part of the same parallel match group as the pairwise match.
> 
> The same also applies if a back-translated all-sides match is subsumed by a back-translated pairwise match - they should be part of the same parallel match group.
> 
> So as each meta-match is back-translated, it creates a corresponding equivalence class to which its back-translated matches belong.
> 
> Once all the meta-matches have been back-translated, the resulting matches are checked for (possibly trivial) partial or full subsumption by matches of greater size; when this occurs the equivalence classes of the subsumed and subsuming matches are unified.
> 
[...]

PR #313, opened 2026-04-02
Parallel match groups drive merging logic

sageserpent-open:
This change addresses Issue #312 by fundamentally shifting the merge strategy from per-file LCS calculations to a flow driven by parallel match groups. These groups represent blocks of code that move together, allowing Kinetic Merge to handle intra- and inter-file moves directly during the alignment phase rather than inferring them later.

The implementation introduces a token-based approach in `SectionedCodeExtension`, where files are partitioned into unmatched sections (gaps) and parallel match groups. A top-level merge aligns these tokens across the base, left, and right sides. Parallel match groups are pre-merged using an inner LCS that preserves internal alignment, and these results are then flattened into the final file merge. This unified treatment simplifies move detection and migration logic, replacing heuristic-based post-processing with a direct mapping from match analysis.

---
*PR created automatically by Jules for task [4462033659096573601](https://jules.google.com/task/4462033659096573601) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
@@ -32,6 +32,8 @@ trait MatchAnalysis[Path, Element]:
 
   def parallelMatchesOnly: MatchAnalysis[Path, Element]
 
+  def parallelMatchGroups: Set[MatchAnalysis.ParallelMatchGroup[Element]]
+
   def reconcileMatches: MatchAnalysis[Path, Element]
 
   def withoutRedundantPairwiseMatches: MatchAnalysis[Path, Element]
@@ -56,6 +58,7 @@ end MatchAnalysis
 
 object MatchAnalysis extends StrictLogging:
   type GenericMatch[Element]    = Match[Section[Element]]
+  type ParallelMatchGroup[Element] = IndexedSeq[GenericMatch[Element]]
   type MatchedSections[Element] =
     MultiDict[Section[Element], GenericMatch[Element]]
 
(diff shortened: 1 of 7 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -2631,34 +2636,167 @@ object MatchAnalysis extends StrictLogging:
...
+          val rootJ = find(j)
+          if rootI != rootJ then metaMatchGroupParent(rootI) = rootJ
+
+        val legitimateMatchesBySection = MultiDict.from(
+          legitimateBackTranslatedParallelMatches.flatMap(m =>
+            m.sections.map(_ -> m)
+          )
+        )
+
+        for
```
sageserpent-open: I've put this comment here, but it is really about the whole union-find / equivalence criterion.

1. I should have stated this before, but let's use something off the shelf rather than write out own union-find implementation (although it seems OK to me, so kudos there). How about the one from Cats: https://typelevel.org/cats-collections/disjointsets.html ?
2. The `overlaps` is nifty in that it is cheap and cheerful, but it relies on the Cartesian product of the matches being enumerated to check for overlaps, which is inefficient. I'm also concerned that because the code sensibly uses a triangular iteration over the Cartesion product, then the testing for overlaps between any given m1 and m2 is only one-way. I'd suggest we enumerate just through all all-sides matches and use `subsumingPairwiseMatchesIncludingTriviallySubsuming` to find any relevant subsuming pairwise matches, including trivially subsuming ones.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -2631,34 +2636,167 @@ object MatchAnalysis extends StrictLogging:
...
+
+            if sharedSidesWithOverlaps >= 2 then
+              union(
+                metaMatchToIndex(matchToMetaMatch(m1)),
+                metaMatchToIndex(matchToMetaMatch(m2))
+              )
+          end for
+        end for
+
+        val metaMatchIndicesByRoot = (0 until metaMatchList.size).groupBy(find)
```
sageserpent-open: Intellij reminds me that we could use `metaMatchList.indices` here, but I'm pushing for a different approach anyway, so this is just an FYI...

sageserpent-open:
This PR has been superseded by work done in the meantime for #312, but I’m keeping it around as a source of inspiration.

For the record, I did come round to Jules’ approach of sticking with a single instance of `MatchesAndTheirSections`, adding an extra field to represent the groups of parallel matches. The representation is different though, and I used Cats’ `DisjointSets` to perform unification of matches that are parallel.

PR #319, opened 2026-04-14
Bolster parallel matches invariant with path consistency check

sageserpent-open:
Modified `MatchAnalysis.scala` to bolster the invariant for parallel matches. The new invariant checks that all matches in a group share consistent file paths for each side. This required adding a refinement step at the end of `reconcileMatches` to split groups that became path-inconsistent due to code motion splits during the reconciliation process. Verified with existing tests, specifically split code motion scenarios.

---
*PR created automatically by Jules for task [12478530646517858524](https://jules.google.com/task/12478530646517858524) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
@@ -1186,6 +1186,99 @@ object MatchAnalysis extends StrictLogging:
         end if
       end purgedOfMatchesWithOverlappingSections
 
+      private def withoutTheseMatches(
+          matches: Iterable[GenericMatch[Element]]
+      ): MatchesAndTheirSections =
+        matches.foldLeft(this) {
+          case (
+                matchesAndTheirSections,
+                allSides @ Match.AllSides(
+                  baseSection,
+                  leftSection,
+                  rightSection
+                )
+              ) =>
+            matchesAndTheirSections.copy(
+              baseSectionsByPath =
+                matchesAndTheirSections.baseExcluding(baseSection),
+              leftSectionsByPath =
+                matchesAndTheirSections.leftExcluding(leftSection),
+              rightSectionsByPath =
+                matchesAndTheirSections.rightExcluding(rightSection),
+              sectionsAndTheirMatches =
+                matchesAndTheirSections.sectionsAndTheirMatches
+                  .remove(baseSection, allSides)
+                  .remove(leftSection, allSides)
+                  .remove(rightSection, allSides),
+              baseFingerprintedInclusionsByPath =
+                matchesAndTheirSections.reinstateInBaseFingerprintedInclusions(
+                  baseSection
+                ),
+              leftFingerprintedInclusionsByPath =
+                matchesAndTheirSections.reinstateInLeftFingerprintedInclusions(
+                  leftSection
+                ),
+              rightFingerprintedInclusionsByPath =
+                matchesAndTheirSections.reinstateInRightFingerprintedInclusions(
+                  rightSection
+                ),
+              parallelMatchesGroupIdsByMatch =
+                parallelMatchesGroupIdsByMatch.removed(allSides)
+            )
+
+          case (
+                matchesAndTheirSections,
+                baseAndLeft @ Match.BaseAndLeft(baseSection, leftSection)
+              ) =>
+            matchesAndTheirSections.copy(
+              baseSectionsByPath =
+                matchesAndTheirSections.baseExcluding(baseSection),
+              leftSectionsByPath =
+                matchesAndTheirSections.leftExcluding(leftSection),
+              sectionsAndTheirMatches =
+                matchesAndTheirSections.sectionsAndTheirMatches
+                  .remove(baseSection, baseAndLeft)
+                  .remove(leftSection, baseAndLeft),
+              parallelMatchesGroupIdsByMatch =
+                parallelMatchesGroupIdsByMatch.removed(baseAndLeft)
+            )
+
+          case (
+                matchesAndTheirSections,
+                baseAndRight @ Match.BaseAndRight(baseSection, rightSection)
+              ) =>
+            matchesAndTheirSections.copy(
+              baseSectionsByPath =
+                matchesAndTheirSections.baseExcluding(baseSection),
+              rightSectionsByPath =
+                matchesAndTheirSections.rightExcluding(rightSection),
+              sectionsAndTheirMatches =
+                matchesAndTheirSections.sectionsAndTheirMatches
+                  .remove(baseSection, baseAndRight)
+                  .remove(rightSection, baseAndRight),
+              parallelMatchesGroupIdsByMatch =
+                parallelMatchesGroupIdsByMatch.removed(baseAndRight)
+            )
+
+          case (
+                matchesAndTheirSections,
+                leftAndRight @ Match.LeftAndRight(leftSection, rightSection)
+              ) =>
+            matchesAndTheirSections.copy(
+              leftSectionsByPath =
+                matchesAndTheirSections.leftExcluding(leftSection),
+              rightSectionsByPath =
+                matchesAndTheirSections.rightExcluding(rightSection),
+              sectionsAndTheirMatches =
+                matchesAndTheirSections.sectionsAndTheirMatches
+                  .remove(leftSection, leftAndRight)
+                  .remove(rightSection, leftAndRight),
+              parallelMatchesGroupIdsByMatch =
+                parallelMatchesGroupIdsByMatch.removed(leftAndRight)
+            )
+        }
+      end withoutTheseMatches
+
       def reconcileMatches: MatchesAndTheirSections =
         val matches = sectionsAndTheirMatches.values.toSet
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1618,6 +1670,44 @@ object MatchAnalysis extends StrictLogging:
...
+
+          val matchesByGroupId =
+            parallelMatchesGroupIdsByMatch.toSeq
+              .groupBy(_._2)
+              .view
+              .mapValues(_.map(_._1))
+
+          matchesByGroupId.foreach { case (groupId, matchesInGroup) =>
+            val basePaths = matchesInGroup.collect {
+              case Match.AllSides(baseSection, _, _)  => baseSources.pathFor(baseSection)
```
sageserpent-open: A minor nitpick - we could extract the `baseSources.pathFor` invocation into a following `.map` that works on the result of the `.collect` call - it would cut down on the repetition slightly.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1618,6 +1670,44 @@ object MatchAnalysis extends StrictLogging:
...
+              case Match.BaseAndLeft(baseSection, _)  => baseSources.pathFor(baseSection)
+              case Match.BaseAndRight(baseSection, _) => baseSources.pathFor(baseSection)
+            }.toSet
+            assert(
+              basePaths.size <= 1,
+              s"Group $groupId has multiple base paths: $basePaths"
```
sageserpent-open: Let's help whoever has to diagnose such a failure and include the set of matches in the assertion diagnostic too.

sageserpent-open:
Apart from minor nitpicks, this looks good to merge. Nevertheless, I would like to explore an alternative approach to ensuring that the final groups of parallel matches are *truly* parallel, as discussed in the chat session with Jules. If that doesn't work out, I'll take this PR (with the nitpicks addressed) as it is.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
sageserpent-open: Good spot, Jules.

sageserpent-open:
Jules, let's go back to the discussion in the chat session, as there may be another conceptual flaw / lurking bug here...

sageserpent-open:
I've updated the assertion failure diagnostics, as well using a more efficient and easier to comprehend order consistency check. Of course, there are still the outstanding test failures...

sageserpent-open:
**NOTE:** I'm happy with having the coloured output from assertions that are furnished by using `fansi.Str` when pretty-printing.

sageserpent-open:
OK, no problem - I've made a new branch to do further work in parallel with this branch off the shared commit fcfc3ecc366ae93ee85b4d99d00fb4c42ecf27b2.

Feel free to resume your work on getting the tests to pass on this branch when you're ready...

sageserpent-open:
> I've switched to a more robust O(N^2) ordering check that correctly handles coincident matches (using <=)

Can you explain the reasoning here? What sort of situation would require this approach?

sageserpent-open:
Good thinking there, Jules.

OK, so if I understand you correctly, you think that the approach in commit fcfc3ec could result in false positives due to potential arbitrary initial match orders prior to (stable) sorting on the offsets on either side.

However, this could only be a problem if multiple matches in the same group shared the same start offset - if a group consists only of parallel matches (as being checked here), then the only way this can happen is that either we have matches of different kinds that have not yet been reconciled, which could be the case if this check was moved out of `reconcilationPostcondition` and into the main `checkInvariant` method.

So we could either incorporate a tiebreaker sort on match kind (or it could be done prior to the sort by offset), or apply a roundtripping sort on both sides prior to doing the final one side versus the other side sorts to put the matches in a canonical order.

I'm keen on keeping the older approach because a) it's more efficient and b) it makes it much easier to see the discrepancies in ordering to the eye.

What do you think?

sageserpent-open:
Actually, I don't think the canonical tiebreaker sorting approach is going to work after all - it could still be the 'wrong order' on the side with colliding start offsets when compared with the other side with distinct start offsets.

It *might* be the case that the string representation of `MatchSynopsis` carries enough information to do the right thing, but this is opaque and brittle.

How about something like this:

```scala
              def checkOrderIsConsistentAcrossSides(
                  firstSide: String,
                  secondSide: String,
                  firstSideStartOffset: GenericMatch[Element] => Option[Int],
                  secondSideStartOffset: GenericMatch[Element] => Option[Int]
              ): Unit =
                def sortBy(
                    startOffsetOf: GenericMatch[Element] => Option[
                      ParallelMatchesGroupId
                    ]
                ): Seq[GenericMatch[Element]] =
                  matchesInGroup
                    .flatMap(aMatch =>
                      startOffsetOf(aMatch)
                        .map(offset => aMatch -> offset)
                    )
                    .sortBy(_._2)
                    .map(_._1)

                // NOTE: in the next two bindings, we have to sort first
                // according to the other side's perspective. This is to avoid
                // situations where several matches happen to have the same
                // start off set on one side but different offsets on the other.
                // As sorting is stable, that could lead to the side with
                // colliding offsets having the 'wrong' initial order that would
                // be preserved by the sort. Tip of the hat to Jules for
                // suggesting the pitfall in the first place.

                val matchesOrderedFromFirstSidePerspective =
                  sortBy(secondSideStartOffset).sortBy(firstSideStartOffset)

                val matchesOrderedFromSecondSidePerspective =
                  sortBy(firstSideStartOffset).sortBy(secondSideStartOffset)

                val commonMatches =
                  matchesOrderedFromFirstSidePerspective.toSet intersect matchesOrderedFromSecondSidePerspective.toSet

                val commonMatchesOrderedFromFirstSidePerspective =
                  matchesOrderedFromFirstSidePerspective.filter(
                    commonMatches.contains
                  )
                val commonMatchesOrderedFromSecondSidePerspective =
                  matchesOrderedFromSecondSidePerspective.filter(
                    commonMatches.contains
                  )
```

It is, I hope, straightforward to see what's going on...

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1618,9 +1618,229 @@ object MatchAnalysis extends StrictLogging:
...
+                  )}""".stripMargin
+              )
+
+              def checkOrderIsConsistentAcrossSides(
+                  firstSide: String,
+                  secondSide: String,
+                  firstSideStartOffset: GenericMatch[Element] => Option[Int],
+                  secondSideStartOffset: GenericMatch[Element] => Option[Int]
+              ): Unit =
+                extension (matches: Seq[GenericMatch[Element]])
```
sageserpent-open: I’m glad you didn’t accept my suggestion verbatim; using an extension allows genuine chaining of the two sort operations - that is easier to read and avoids a nasty bug where irrelevant matches could creep into one side’s ordering result.

Nice catch.

sageserpent-open:
I can’t merge this PR to branch `main` by virtue of its success in achieving its objectives - it’s found new test failures! 😁 

Nevertheless it’s a good job, so I’ll merge it on to the other feature branch to go forward.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -694,196 +732,237 @@ object MatchAnalysis extends StrictLogging:
...
+        private def eatIntoSection(
+            firstSide: Sources[Path, Element],
+            firstSection: Section[Element],
+            secondSide: Sources[Path, Element],
+            secondSection: Section[Element],
+            groupIdsBySortedBiteEdge: Map[BiteEdge, collection.Set[
+              ParallelMatchesGroupId
+            ]]
+        ): Vector[
+          (Section[Element], Section[Element], Set[ParallelMatchesGroupId])
```
sageserpent-open: Something that just occurred to me: if we pass in a lambda the constructs a PairwiseMatch from a pair of sections from either side to `eatIntoSection`, then this function can return pairwise matches directly; that simplifies the caller’s logic and hopefully avoids the repeated analysis of the flavour of the pairwise match.

In fact, would it make sense to pass the large pairwise match being eaten into to this function - and thus rename it accordingly?

The alternative is to leave things as they are with the assembly of the pairwise matches being done in the caller, in which case I suppose the function should be renamed to `eatIntoSections`…

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -694,196 +732,237 @@ object MatchAnalysis extends StrictLogging:
...
+                        leftSection,
+                        rightSources,
+                        rightSection,
+                        groupIdsBySortedBiteEdge
+                      )
+                  end triples
+
+                  val fragments = triples.map {
+                    case (firstSection, secondSection, _) =>
+                      (pairwiseMatch match
```
sageserpent-open: It would be nice to avoid the repeated analysis of the flavour of pairwise match here…

sageserpent-open:
This looks good to me, nice work.

There is a minor nitpick in that the commit 35f0436c is really a merge from commit 7286d0c2 with the `theirs` strategy, but has been done as a cut-and-paste of content.

I'll try rebuilding the last three commits as a rebase to preserve the original authorship, but other than that, it's ready.

sageserpent-open:
> There is a minor nitpick in that the commit [35f0436](https://github.com/sageserpent-open/kineticMerge/commit/35f0436caa2dd2a6a35c86e941f3123c76152fb2) is really a merge from commit [7286d0c](https://github.com/sageserpent-open/kineticMerge/commit/7286d0c2549022fb08298d56808adcd1b8ef35fb) with the `theirs` strategy, but has been done as a cut-and-paste of content.

Strictly speaking 35f0436 was a squashed merge, only without the extra parent of 7286d0c. I've simply cherry-picked the following commits from Jules on to 7286d0c and force-pushed to update this branch.

Issue #266, opened 2025-12-10
Finesse Challenging Merge


sageserpent-open:
Attempt to merge two branches in the repository for **Kinetic Merge** with Kinetic Merge 1.12.0.

These branches are the JJ integration spike delivered in #136 (c080623403b4322069505eac40ff34a612b6c314) and the mainline as of just after v1.12.0 (68118e57bc3be78cf4b4727faed5bad8122a23ca).

The JJ integration spike substantially rewrote the driver application `Main`, stripping out all interaction with Git porcelain and plumbing and replacing it where appropriate with reading and writing files from three directories supplied as command line arguments and pre-populated with the input source trees for the base, left and right of the merge. The outputs of the merge are written back into all three directories.

Meanwhile, work on the mainline has added in support for binary files and some exploratory work for submodule support, as well as some finessing of reporting file renames versus file content transplants. Some refactoring of logic was done in there, and a lot of logic was duplicated and modified to account for situations where the same path on different sides of the merge might vary between being binary or textual in content.

Back on the JJ spike, nearly all of the tests in `MainTest` were patched up to drive the severely modified form of `Main` in an appropriate manner, while the mainline has modified the test suite, adding new tests and possibly changing some existing ones a bit too.

sageserpent-open:
Ok, let's try core Git merge then:

```bash
git merge --no-commit c080623 68118e5
Auto-merging src/test/scala/com/sageserpent/kineticmerge/MainTest.scala
CONFLICT (content): Merge conflict in src/test/scala/com/sageserpent/kineticmerge/MainTest.scala
Auto-merging src/main/scala/com/sageserpent/kineticmerge/Main.scala
CONFLICT (content): Merge conflict in src/main/scala/com/sageserpent/kineticmerge/Main.scala
Automatic merge failed; fix conflicts and then commit the result.
```

OK, two conflicted files, so surprises there.

Looking through `MainTest.scala`, we see conflicts like:

```scala
<<<<<<< ours
              if noCommit then
                verifyAConflictedOrNoCommitMergeDoesNotMakeACommitAndLeavesADirtyIndex(
                  path
                )(
                  flipBranches,
                  commitOfConcurrentlyDeletedFileBranch,
                  commitOfMasterBranch,
                  ourBranch,
                  exitCode
                )
              else
                verifyMergeMakesANewCommitWithACleanIndex(path)(
                  commitOfConcurrentlyDeletedFileBranch,
                  commitOfMasterBranch,
[...]

PR #322, opened 2026-04-22
Approximate per-file LCS using parallel match groups

sageserpent-open:
Implements issue #321: Use parallel matches groups to generate per-file longest common subsequences. This optimization avoids full DP on sections by aligning at the block (match group) level first.

---
*PR created automatically by Jules for task [11265001918498214285](https://jules.google.com/task/11265001918498214285) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
@@ -58,6 +58,8 @@ trait MatchAnalysis[Path, Element]:
 
   def sectionsAndTheirMatches: MatchedSections[Element]
 
+  def parallelMatchesGroupIdsByMatch: MultiDict[GenericMatch[Element], Int]
+
   def matches: Set[GenericMatch[Element]] =
     sectionsAndTheirMatches.values.toSet
 end MatchAnalysis
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
@@ -12,7 +12,8 @@ import com.google.common.hash.{Funnel, HashFunction, PrimitiveSink}
 import com.sageserpent.kineticmerge
 import com.sageserpent.kineticmerge.core.MatchAnalysis.{
   GenericMatch,
-  MatchedSections
+  MatchedSections,
+  ParallelMatchesGroupIdsByMatch
 }
 import com.sageserpent.kineticmerge.{
   NoProgressRecording,
@@ -58,7 +59,7 @@ trait MatchAnalysis[Path, Element]:
 
   def sectionsAndTheirMatches: MatchedSections[Element]
 
-  def parallelMatchesGroupIdsByMatch: MultiDict[GenericMatch[Element], Int]
+  def parallelMatchesGroupIdsByMatch: ParallelMatchesGroupIdsByMatch[Element]
 
   def matches: Set[GenericMatch[Element]] =
     sectionsAndTheirMatches.values.toSet
@@ -68,6 +69,11 @@ object MatchAnalysis extends StrictLogging:
   type GenericMatch[Element]    = Match[Section[Element]]
   type MatchedSections[Element] =
     MultiDict[Section[Element], GenericMatch[Element]]
+
+  type ParallelMatchesGroupId = Int
+
+  type ParallelMatchesGroupIdsByMatch[Element] =
+    MultiDict[GenericMatch[Element], ParallelMatchesGroupId]
 
   /** Analyse matches between the sources of {@code base}, {@code left} and
     * {@code right}.
@@ -188,10 +194,8 @@ object MatchAnalysis extends StrictLogging:
 
     type FingerprintedInclusions = Diet[Int]
 
-    type ParallelMatchesGroupId = Int
-
     type ParallelMatchesGroupIdsByMatch =
-      MultiDict[GenericMatch[Element], ParallelMatchesGroupId]
+      MatchAnalysis.ParallelMatchesGroupIdsByMatch[Element]
 
     val tiebreakContentSamplingLimit = 5
 
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
@@ -101,154 +101,80 @@ object SectionedCode extends StrictLogging:
         )
 
       Right(new SectionedCode[Path, Element]:
-        private val lcsByPath: Map[Path, LongestCommonSubsequence[Section[
-          Element
-        ]]] =
-          val parallelMatchesGroupIdsByMatch =
-            matchesAndTheirSections.parallelMatchesGroupIdsByMatch
-          val tinyParallelMatchesGroupIdsByMatch =
-            tinyMatchesAndTheirSectionsOnly.parallelMatchesGroupIdsByMatch
+        private val lcsByPath: mutable.Map[Path, LongestCommonSubsequence[
+          Section[Element]
+        ]] = mutable.Map.empty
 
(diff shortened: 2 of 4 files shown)


sageserpent-open:
I'm taking a look at this locally, was running the manual benchmark example discussed in #266.

That benchmark performs a pathological merge in the Kinetic Merge repository itself (the merge isn't intended to be committed, it is purely to serve as a challenging example), merging *from* commit 55fef2f785334b7bf99fc592dd101a462cbb2c6d *to* 6e0b6821fa8fe53465c3355793b61ac425a10c1a:

```bash
sbt packageExecutable

git checkout 6e0b6821fa8fe53465c3355793b61ac425a10c1a

target/kinetic-merge --minimum-match-size=2 --minimum-ambiguous-match-size=2  --no-commit --ambiguous-matches-threshold=50 55fef2f785334b7bf99fc592dd101a462cbb2c6d         

```

The specific tweaks to the command line flags vary in that ticket, but it serves as a useful benchmark.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -58,6 +58,8 @@ trait MatchAnalysis[Path, Element]:
 
   def sectionsAndTheirMatches: MatchedSections[Element]
 
+  def parallelMatchesGroupIdsByMatch: MultiDict[GenericMatch[Element], Int]
+
```
sageserpent-open: We should hoist `ParallelMatchesGroupIdsByMatch` up into the companion object of `MatchAnalysis` (as well as `ParallelMatchesGroupId`), so that the abstract and concrete definitions of this property look consistent.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
```diff
@@ -93,6 +101,80 @@ object SectionedCode extends StrictLogging:
         )
 
       Right(new SectionedCode[Path, Element]:
+        private val lcsByPath: mutable.Map[Path, LongestCommonSubsequence[
```
sageserpent-open, line 105: I think this is a cache for on-demand computation of an LCS? Let's rename the private val to emphasize that (if I've understood correctly).

sageserpent-open:
I think there has been a misunderstanding here: commit 1df8dee383a3737a39a3568d90817e63a001048f seems to attempt to implement the idea from #321, although it breaks tests.

The following commit 138bb44e91eeae0264d522858124d0aa9e9270df abandons the use of blocks to generate an approximate LCS at section-level and goes back to computing section-level LCSes directly using dynamic programming - so we're back to the extreme performance hit from doing that.

It does cache the LCSes, but in reality, these are only computed once for each file (or splice) anyway, so cacheing doesn't buy us anything.

I shall take a look at 1df8dee383a3737a39a3568d90817e63a001048f to see what can be used there...

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
sageserpent-open: Actually, given what I've said elsewhere about the overall approach being wrong in commit 138bb44e91eeae0264d522858124d0aa9e9270df, this doesn't really matter anymore.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
```diff
@@ -93,6 +102,165 @@ object SectionedCode extends StrictLogging:
...
+              )(using
+                Eq[Block],
+                blockSized,
+                com.sageserpent.kineticmerge.NoProgressRecording
+              )
+
+              def expand(
+                  contributions: IndexedSeq[Contribution[Block]]
+              ): IndexedSeq[Contribution[Section[Element]]] =
+                contributions.flatMap {
```
sageserpent-open, line 237: This puzzles me - I had envisaged that the expansion of a block back to sections would have to respect both the block-level contribution kind *and* whether each section in the block was involved as an all-sides match, pairwise match or was unmatched.

So given, say a `Common` contribution at the block level, then a section in it participating in an all-sides match would give rise to a `Common` contribution at the section level, whereas a section participating in say, a `BaseAndLeft` pairwise match within the same block would contribute a mere `CommonToBaseAndLeftOnly` at the section level. An unmatched section in that block would contribute just a `Difference`.

If the contribution at the block level is one of the `CommonToBaseAndLeftOnly`, etc, then we don't expect to see its sections participate in all-sides matches and thus the section level contributions are restricted to just `CommonToBaseAndLeftOnly`, etc and `Difference`.

If the contribution at the block level is a `Difference` then all of the sections contribute a `Difference`.

I thought this was absolutely vital to get fidelity with the old section-level LCS dynamic programming computation; nevertheless it seems to work anyway as it currently is in this PR.

Perhaps we're missing some important tests, or have I failed to understand this correctly?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
```diff
@@ -93,6 +102,165 @@ object SectionedCode extends StrictLogging:
...
+
+              val baseBlocks  = compress(baseSections)
+              val leftBlocks  = compress(leftSections)
+              val rightBlocks = compress(rightSections)
+
+              given blockSized: Sized[Block] = _.sections.map(_.size).sum
+
+              given Eq[Block] with
+                override def eqv(x: Block, y: Block): Boolean =
+                  x.sections.size == y.sections.size && {
```
sageserpent-open, line 211: I was expecting this to only have to check on having a shared group id and nothing more - indeed I tried modifiying the definition to:

```scala
              given Eq[Block] with
                override def eqv(x: Block, y: Block): Boolean =
                  x.groupIds.intersect(y.groupIds).nonEmpty
              end given
```

and it still passes all the tests.

sageserpent-open: I'm not suprised this isn't been caught by the tests - it turns out that `SectionedCode.lcs` *isn't* being called - the old approach to computing the LCS is still being used.

sageserpent-open:
Jules, I've pushed a commit outlining a change in direction for the implementation.

See the notes I wrote:

https://github.com/sageserpent-open/kineticMerge/issues/321#issuecomment-4319518433

https://github.com/sageserpent-open/kineticMerge/issues/321#issuecomment-4319818815

sageserpent-open:
Jules, I've pushed another commit detailing the latest change of plan.

On reflection I think my idea of doing the LCS dynamic programming calulation using just group ids is a step too far - it ignores the sizes of the content covered by the blocks implied by the group ids, and also doesn't include representation of the content that lies in-between these implied blocks. These should factor into the LCS computation.

What I'm now proposing is a compromise between the original approach and the group id approach - so it's back to constructing blocks that cover all the sections on a given side that are encompassed by a a group of parallel matches, only this time, the block either has a single group id or none at all. If multiple groups of parallel matches cover the same content on a given side, that results in multiple blocks that happen to have the saem sequence of sections, but different groups ids. The block ordering is amended to work in terms of triple of the start offset of the leading matched section, the negative of the size of the covered content and then as a final tiebreaker, the group id.

I've added an implementation plan inside `SectionedCodeExtension.longestCommonSubsequenceOf` with more detail. Hopefully that is enough to get us unblocked on this ...

sageserpent-open:
Again, the latest commit has reverted back to the state where `SectionedCodeExtension` directly calls the legacy dynamic programming algorithm to calculate the section-level LCS for a path.

Furthermore, my suggestions haven’t been acted on: so the unused code that tries to implement this ticket is still much of not completely the same as a few commits ago.

We’re going round in circles for some reason. This is all the more puzzling, because the chat session with Jules seems to show that it has taken heed of both the PR comments and the steer I have it within the chat session itself.

I have made a branch off commit a9a63a5dc47ea66e007dec69ab7585be24346ed6 and am doing this myself; using what Jules had written as a guide for the code structure.

PR #329, opened 2026-05-09
Restore block-LCS optimization and add merge smoke tests

sageserpent-open:
This submission restores the block-level Longest Common Subsequence (LCS) optimization in `SectionedCodeExtension`, which significantly improves merge performance for large files. It incorporates improved diagnostics for `blocksFrom` contract violations and introduces a robust property-based smoke test in `SectionedCodeTest` that exercises the full merge workflow. Additionally, it fixes a bug where the merge process would fail with a `NoSuchElementException` if a file was added or deleted on one of the merge sides.

---
*PR created automatically by Jules for task [6994334355823168847](https://jules.google.com/task/6994334355823168847) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
@@ -40,11 +40,10 @@ case class FirstPassMergeResult[Element](
     rightPreservations: Set[Element]
 )
 
-enum Side:
+enum MoveDestinationSide:
   case Left
-  case Base
   case Right
-end Side
+end MoveDestinationSide
 
 object FirstPassMergeResult:
   opaque type Recording[Element] =
--- a/src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
@@ -42,6 +42,7 @@ case class FirstPassMergeResult[Element](
 
 enum Side:
   case Left
+  case Base
   case Right
 end Side
 
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
@@ -29,7 +29,7 @@ import com.typesafe.scalalogging.StrictLogging
 import monocle.syntax.all.*
 
 import scala.annotation.tailrec
-import scala.collection.immutable.{MultiDict, SortedMap}
+import scala.collection.immutable.MultiDict
 import scala.collection.{IndexedSeqView, Searching}
 import scala.math.Ordering.Implicits.seqOrdering
 
@@ -54,10 +54,14 @@ object SectionedCodeExtension extends StrictLogging:
         .flatMap(sectionedCode.parallelMatchesGroupIdsByMatch.get)
 
       case class Block(
-          parallelMatchesGroupId: Option[ParallelMatchesGroupId],
+          parallelMatchesGroupId: ParallelMatchesGroupId,
           sectionsCoveredByGroup: IndexedSeq[Section[Element]]
       ):
         require(sectionsCoveredByGroup.nonEmpty)
+        sectionsCoveredByGroup.zip(sectionsCoveredByGroup.tail).foreach {
+          case (predecessor, successor) =>
+            require(predecessor.onePastEndOffset == successor.startOffset)
+        }
 
         // TODO: do we actually need the sections at all if we've got these? Why
         // not just store an offset interval?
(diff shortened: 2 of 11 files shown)


src/test/scala/com/sageserpent/kineticmerge/core/SectionedCodeTest.scala
```diff
@@ -1918,6 +1927,356 @@ class SectionedCodeTest:
...
+                  end for
+                end if
+            }
+        )
+      }
+
+    testPlans
+      .withStrategy(caseSupplyCycle =>
+        if caseSupplyCycle.isInitial then
+          CasesLimitStrategy.timed(Duration.apply(1, TimeUnit.MINUTES))
```
sageserpent-open: This duration is too low. I tried running this test locally via IntelliJ and saw test cases that failed: I suspect the VM used to run this test in the Jules environment is slower, so it doesn't encounter the first failure in the time available. I suggest raising this to 2 minutes at least to flush out the failure....

src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
```diff
@@ -42,6 +42,7 @@ case class FirstPassMergeResult[Element](
 
 enum Side:
   case Left
+  case Base
```
sageserpent-open: This enumeration shouldn't be changed, because it deliberately models the two sides in a merge that have the *changes*, so `Base` should not go here. I suggest either defining a local enumeration within `SectionedCodeExtension.longestCommonSubsequenceOf`, or finding a way of generalising the logic in `assignContributionsOnOneSide` so that the side-specific tweaks to the logic can be passed in as an abstracted trait. Or maybe just duplicate `assignContributionsOnOneSide` for each side-specific tweak!

Furthermore, the name of *this* enumeration should be changed to `ChangedSide` to avoid this kind of mistake in the future - that was a poor choice of name on my part in the past.

src/test/scala/com/sageserpent/kineticmerge/core/SectionedCodeTest.scala
```diff
@@ -1918,6 +1927,356 @@ class SectionedCodeTest:
     assert(6 == pairwise.size)
   end parallelMatches
 
+
+  @TestFactory
+  def mergeSmokeTest(): DynamicTests =
+    case class TestPlan(
```
sageserpent-open: This has been cut and paste from `SectionedCodeTest.matchingSectionsAreFound`, but I asked in the chat session that this be hoisted to avoid this duplication. Can we do that, please.

sageserpent-open:
As a general comment about the PR, I asked in the chat session for this to be based off commit 362f3400cdab092318dddc959242b99d6f356985.

What has happened is that the branch has been started directly off `main`, but incorporates the changes not already upstream that lead up to 362f3400cdab092318dddc959242b99d6f356985 as duplicates, rather than as part of the history of this branch. That is OK from the point of view of getting code to work, but it makes attribution difficult and loses some the insight as to why the changes were made.

For now, let's carry on with the branch as it is, but later on I shall rebase it on to 362f3400cdab092318dddc959242b99d6f356985 to tidy up the history and isolate Jules' own changes.

sageserpent-open:
@jules - there are no changes in the last commit 4140a404ca401f1e743eaf3e20bfa54d8eab67bf that you pushed out. Did you stage your latest changes correctly?

Therefore I'm seeing failures in `mergeSmokeTest`.

However, there is some 'good' news too - as was set out in the chat session, the primary purpose of the new test was to flush out the block id assertion failure mentioned here: https://github.com/sageserpent-open/kineticMerge/issues/321#issuecomment-4397415660.

This was sidetracked by the LCS construction synchronization failure that we encountered and you have already fixed on the way, so this is good progress, now we can see the original failure and work on that.

sageserpent-open:
More encouraging news - as of commit 650ea4f3fba46062b5895d7716ba5bdad94e4fb2, building Kinetic Merge and retrying it as per https://github.com/sageserpent-open/kineticMerge/issues/321#issuecomment-4397835226 now avoids the second assertion failure, presumably due to the changes made to `MergeResult` in this PR.

So it looks like we're down to just the block is reuse issue now. I'm going to take a quick look at this myself in tandem...

sageserpent-open:
I have force-pushed the rebase of this branch to clarify the authorship and tease out Jules' work from my own. I did work on extending the tests to include a merge smoke test myself, so there are some changes there, but nothing of any significance - the new test `SectionedCodeTest.mergeSmokeTest` is more or less the same as before the rebase.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -30,17 +38,648 @@ import scala.collection.{IndexedSeqView, Searching}
...
+              .slice(startingSectionIndex, 1 + endingSectionIndex)
+              .map(section => section -> groupIdsOf(section))
+          )
+
+          s"""
+             |$prelude
+             |
+             |Group id: $groupId
+             |
+             |Sections in group:
```
sageserpent-open: I should have made this clearer in the diagnostic - where it says "Sections in group:", it should really be "Matched sections in group:".

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -30,17 +38,648 @@ import scala.collection.{IndexedSeqView, Searching}
...
+          ): @unchecked
+
+          val Searching.Found(endingSectionIndex) = file.searchByStartOffset(
+            finalSection
+              .getOrElse(sectionsFromGroupOnRelevantSide.last)
+              .startOffset
+          ): @unchecked
+
+          val allSectionsBetweenBracketsDiagnostic = pprintCustomised(
+            file.sections
```
sageserpent-open: It would be handy to know which group each section belongs to here, so we can see what is broken up a single block into multiple pieces separated by 'intruder' sections.

sageserpent-open: Not only does this code already do that, but I was the author. Doh! :facepalm:

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
sageserpent-open: No longer needed as the blocks are constructed directly from groups rather than by scanning paths.

src/main/scala/com/sageserpent/kineticmerge/core/FirstPassMergeResult.scala
sageserpent-open: No longer relevant as the blocks are constructed directly from groups rather than by scanning paths. I have changed the name as this enumeration to `MoveDestinationSide` to make its intent clearer (it is always used for that purpose), and I've dropped the `Base` case.

sageserpent-open:
Jules, I've implemented a first cut of constructing the blocks up front directly from the groups, as per our discussion in the chat, so that resolves my earlier comments.

This means that `SectionsAndTheirMatches.mergeSmokeTest` now passes, which is good - but there are a few tests that are broken due to block construction failing; this is due to attempts to look up a section belonging to a block failing because the section is not present in the final file breakdown for the given side and path.

I suspect that the cleaning up of overlapping sections is the culprit here - the groups are computed earlier on, so possibly a group has a match that has a section that has been cleaned up...

PR #340, opened 2026-05-28
Fix duplicate block merge demotion and prioritize all-sides matches

sageserpent-open:
This change addresses the issue where duplicate blocks on different sides of a merge would cause matches to be discarded, resulting in content being incorrectly marked as a difference.

Key changes:
1.  **Match Demotion:** The merge logic in `SectionedCodeExtension.scala` now allows multiple matches from a group of sharing sections. The first all-sides match is prioritized, and subsequent matches that share sections with already-seen matches are demoted to pairwise matches or omitted if they have no unique sections left.
2.  **All-Sides Match Prioritization:** Following code review feedback, the logic was refined to explicitly partition and process all-sides matches first, ensuring they are not "starved" of sections by pairwise matches in the same group.
3.  **Expanded Testing:** A new test case for triplication was added to `BlockDuplicationAndCondensationTests.scala` to ensure the logic robustly handles higher degrees of duplication.
4.  **Ambiguity Handling:** The `ambiguousMatchesThreshold` was increased in the triplication test to allow the match analysis phase to identify all relevant duplicate segments.

---
*PR created automatically by Jules for task [2014834485902875714](https://jules.google.com/task/2014834485902875714) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
@@ -385,16 +385,10 @@ object SectionedCodeExtension extends StrictLogging:
             )
           )
 
-          val unionStep = previouslySeenMatchesSharingAtLeastOneSection.foldLeft(
-            State.pure[DisjointSets[Match[Section[Element]]], Unit](())
-          )((step, previouslySeenMatch) =>
-            step >> DisjointSets.union(
-              matchJustEncountered,
-              previouslySeenMatch
-            ).void
-          )
-
-          unionStep >> resultStep
+          previouslySeenMatchesSharingAtLeastOneSection.headOption
+            .fold(ifEmpty = resultStep)(
+              DisjointSets.union(matchJustEncountered, _) >> resultStep
+            )
         } >> DisjointSets.toSets
 
         // NOTE: this is a tiebreaking order that works across from base to left
@@ -409,10 +403,7 @@ object SectionedCodeExtension extends StrictLogging:
           (
             aMatch.baseContribution.map(_.startOffset),
             aMatch.leftContribution.map(_.startOffset),
-            aMatch.rightContribution.map(_.startOffset),
-            aMatch match
-              case _: Match.AllSides[Section[Element]] => 0
-              case _                                   => 1
+            aMatch.rightContribution.map(_.startOffset)
           )
         )
 
--- a/src/test/scala/com/sageserpent/kineticmerge/core/BlockDuplicationAndCondensationTests.scala
+++ b/src/test/scala/com/sageserpent/kineticmerge/core/BlockDuplicationAndCondensationTests.scala
@@ -52,7 +52,7 @@ class BlockDuplicationAndCondensationTests:
       minimumMatchSize = 1,
       thresholdSizeFractionForMatching = 0,
       minimumAmbiguousMatchSize = 0,
-      ambiguousMatchesThreshold = 4
+      ambiguousMatchesThreshold = 10
     )
 
     Trials.api.booleans.withLimit(2).dynamicTests { mirrorImage =>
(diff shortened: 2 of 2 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -384,16 +385,22 @@ object SectionedCodeExtension extends StrictLogging:
             )
           )
 
-          previouslySeenMatchesSharingAtLeastOneSection.headOption
-            .fold(ifEmpty = resultStep)(
-              DisjointSets.union(matchJustEncountered, _) >> resultStep
-            )
+          val unionStep = previouslySeenMatchesSharingAtLeastOneSection.foldLeft(
```
sageserpent-open: I like the way that using a `.foldLeft` sidesteps the duplicated use of `resultStep` in the original code.

The use of `void` is a useful tip, thanks.

One thing - the union operation is transitive: given that `DisjointSets` has the usual optimisations such as path compression and depth balancing, do we need to left-fold over *all* of `previouslySeenMatchesSharingAtLeastOneSection`?

If it's that important to avoid the duplicated use of `resultStep`, how about `previouslySeenMatchesSharingAtLeastOneSection.headOption.iterator.foldLeft` or something like that?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -402,7 +409,10 @@ object SectionedCodeExtension extends StrictLogging:
           (
             aMatch.baseContribution.map(_.startOffset),
             aMatch.leftContribution.map(_.startOffset),
-            aMatch.rightContribution.map(_.startOffset)
+            aMatch.rightContribution.map(_.startOffset),
+            aMatch match
+              case _: Match.AllSides[Section[Element]] => 0
+              case _                                   => 1
```
sageserpent-open: Is this strictly necessary given that demotion only applies to all-sides matches? The all-sides matches are *partitioned* in `representativeMatchesFrom`, so does the more precise tiebreaking ordering matter here?

If I'm wrong, please explain...

src/test/scala/com/sageserpent/kineticmerge/core/BlockDuplicationAndCondensationTests.scala
```diff
@@ -435,4 +436,98 @@ class BlockDuplicationAndCondensationTests:
...
+      println(
+        s"The other side with triplication contributions: ${pprintCustomised(contributionsOnTheOtherSideWithTriplication)}"
+      )
+
+      assert(
+        Vector(Contribution.Common(blockContent)) == baseContributions
+          .map(_.map(_.content))
+      )
+
+      val expectedTriplicatedSide = Vector(
```
sageserpent-open, line 515: Nice - we should do this in the duplication tests too, good idea.

sageserpent-open (commented):
Just a couple of points in this review, but overall this looks good to me, nice work.

src/test/scala/com/sageserpent/kineticmerge/core/BlockDuplicationAndCondensationTests.scala
```diff
@@ -435,4 +436,98 @@ class BlockDuplicationAndCondensationTests:
     }
   end overlappingBlocksAreSeparatedOnOneSide
 
+  @TestFactory
+  def aBlockIsTriplicatedOnTwoSides(): DynamicTests =
+    val configuration = Configuration(
+      minimumMatchSize = 1,
+      thresholdSizeFractionForMatching = 0,
+      minimumAmbiguousMatchSize = 0,
+      ambiguousMatchesThreshold = 10
```
sageserpent-open, line 445: Good catch!

Issue #339, opened 2026-05-28
Fix failure of `BlockDuplicationAndCondensationTests#overlappingBlocksAreSeparatedOnOneSide`


sageserpent-open:
Need a better title - this ticket is a work in progress.

Test `BlockDuplicationAndCondensationTests#overlappingBlocksAreSeparatedOnOneSide` was disabled in commit 39673c8612af8869813ed5846d34330934c84f57.

Why this has been done, and what to do about it, along with a related problem involving tiny matches is the subject of this ticket.

sageserpent-open:
Why was `BlockDuplicationAndCondensationTests#overlappingBlocksAreSeparatedOnOneSide` disabled?

This was because of two things:
1. The test doesn't actually cause the creation of two overlapping blocks, because the match analysis rejected the overlapping matches that would have led to the construction of overlapping blocks. It is supposed to do that, duh!
2. Unfortunately that leads to none of the content being matched, so for the current approach to building a section level LCS, everything will be treated as a difference, so the test fails.

Now it is possible (and it should be done) to reformulate the test setup by inserting bystander content between the test's leading, overlapped and trailing content - the bystander content is devised so that it won't match, and this forces smaller matches that lead to the desired overlapping blocks. So now the blocks can cover unmatched content too, which is already supported.

However, I don't want to sweep the problem with the current formulation of the test under the carpet - it shouldn't fail anyway.

sageserpent-open:
Let's deal with the second thing here and raise another ticket for the first after some discussion.

As an aside, there is logic in `MergeResult` that can clean up a spurious conflict where there is agreement across the sides, so this isn't necessarily a problem in the downstream processing, but the LCS shouldn't exhibit this behaviour - it always used to be able to find 'last minute matches' as it aligned elements, and that is what is missing from the new approach.

What's gone wrong in the new approach to computing a section-level LCS is that unmatched filler sections that aren't covered by blocks aren't actually merged at all - they are demoted directly to difference contributions. The section-level merging only takes place within blocks, and while that will do last-minute matching for filler sections that happen to be covered by a block (because they are flanked by the block's matched sections), any filler sections outside will be summarily demoted, even if they could actually match each other.

The old approach to constructing an LCS ploughed through all of the sections, so nothing could get away.

sageserpent-open:
[...]

PR #344, opened 2026-05-31
Implement last-minute matches for filler sections (#339)

sageserpent-open:
This change addresses the "second thing" from issue #339. By representing unmatched gaps as 'filler blocks' within the `SectionedCode` structure, we allow the merge engine's block-level LCS to align these regions based on their actual content. This facilitates "last-minute matches" during the section-level merge refinement, even for sections that were not part of any parallel match group identified in the initial analysis. The `Eq[Block]` implementation was updated to fallback to content comparison when group IDs are missing, and `SectionedCode.of` now ensures total file coverage by blocks.

---
*PR created automatically by Jules for task [14196552909955845735](https://jules.google.com/task/14196552909955845735) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
@@ -161,33 +161,36 @@ object SectionedCode extends StrictLogging:
 
           val sortedMatchedBlocks = matchedBlocks.sortBy(_.startOffset)
 
-          val fillerBlocks = mutable.Buffer.empty[Block[Element]]
-          var currentEnd   = 0
-          for block <- sortedMatchedBlocks do
-            if block.startOffset > currentEnd then
+          val (onePastLastEndOffset, fillerBlocks) =
+            sortedMatchedBlocks.foldLeft(0 -> Vector.empty[Block[Element]]) {
+              case ((currentEnd, fillers), block) =>
+                val nextFillers =
+                  if block.startOffset > currentEnd then
+                    val Searching.Found(startingSectionIndex) =
+                      file.searchByStartOffset(currentEnd): @unchecked
+                    val Searching.Found(endingSectionIndex) =
+                      file.searchByStartOffset(block.startOffset): @unchecked
+                    fillers :+ Block(
+                      parallelMatchesGroupId = None,
+                      sectionsCoveredByGroup = file.sections
+                        .slice(startingSectionIndex, endingSectionIndex)
+                    )
+                  else fillers
+                (currentEnd max block.onePastEndOffset) -> nextFillers
+            }
+
+          val allFillerBlocks =
+            if onePastLastEndOffset < file.size then
               val Searching.Found(startingSectionIndex) =
-                file.searchByStartOffset(currentEnd): @unchecked
-              val Searching.Found(endingSectionIndex) =
-                file.searchByStartOffset(block.startOffset): @unchecked
-              fillerBlocks += Block(
+                file.searchByStartOffset(onePastLastEndOffset): @unchecked
+              fillerBlocks :+ Block(
                 parallelMatchesGroupId = None,
                 sectionsCoveredByGroup =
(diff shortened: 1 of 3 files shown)


sageserpent-open:
This commit is a squashed merge of work done on branch `issue-339-last-minute-matches`.

In order to preserve attribution, I've force-pushed that branch to this one, and credited Jules as the co-author on commit d8ebfd740c7cfb487a7c8cda84db73e733d7ec37 which was a code-drop of Jules' uncommitted work in progress.

The content of the force-pushed branch head is the same as that originally raised in this PR.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -78,8 +79,19 @@ object SectionedCodeExtension extends StrictLogging:
...
         Eq.or(
-          Eq.by(_.parallelMatchesGroupId),
-          Eq.by(block => groupsOfParallelMatches(block.parallelMatchesGroupId))
+          (lhs, rhs) =>
+            (
+              lhs.parallelMatchesGroupId,
+              rhs.parallelMatchesGroupId
+            ) match
+              case (Some(lhsGroupId), Some(rhsGroupId)) =>
+                lhsGroupId == rhsGroupId
```
sageserpent-open: Jules, you originally wrote:

```scala
lhsGroupId == rhsGroupId || groupsOfParallelMatches(
                  lhsGroupId
                ) == groupsOfParallelMatches(rhsGroupId)
```

here. That wasn't right, because it pulled in the intrinsic `Eq[Match[Section[Element]]]` defined by parallel matches groups that doesn't work across different matches groups or for matches in general. 

I fixed that in commit fb46b16a74e6161097d54a48f5cc5fd8404d2057, but I think my fix was too draconian - the original code on the baseline for this branch read as:

```scala
Eq.or(
          Eq.by(_.parallelMatchesGroupId),
          Eq.by(block => groupsOfParallelMatches(block.parallelMatchesGroupId))
        )
```
and that picked up the correct `Eq[Match[Section[Element]]]` defined locally just before this code snippet. That meant that if two blocks were associated with distinct parallel matches groups and they agreed purely on the matches (and therefore allowed some leeway for filler sections within the blocks), then they would still be deemed equivalent.

I think you did the right thing having the extra subclause, but used the wrong equality.

You should reinstate the extra subclause here, but recast it as:

```scala
lhsGroupId == rhsGroupId || Eq.eqv(groupsOfParallelMatches(lhsGroupId), groupsOfParallelMatches(rhsGroupId))
```

Probably worth updating the preceding comment to avoid repeating this mistake, given that it's caught us both out so far.

sageserpent-open (commented):
Overall, looks good to me, but as I stated in the review comment, I think I overcooked the fix for the outstanding test failure.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCode.scala
```diff
@@ -144,25 +144,57 @@ object SectionedCode extends StrictLogging:
...
           .map(blockFrom)
           .collect { case Some((path, block)) => path -> block }
           .groupMap(_._1)(_._2)
-          .map((path, blocks) =>
-            path -> blocks.toIndexedSeq.sortBy(block =>
-              (
-                block.startOffset,
-                block.onePastEndOffset,
-                block.parallelMatchesGroupId
+
```
sageserpent-open, line 157: I wonder, would a pure functional approach read better here?

Something similar to this is done in `MappedContentSources.filesByPathUtilising` to add filler sections in between the matched mandatory sections - maybe it could be adapted?

Alternatively, sort the blocks with group ids to start with, work up in pairs via the `<blocks>.zip(<blocks>.tail)` approach and look at the gaps between them, filtering out empty or reversed gaps to create the filler blocks. A merge of two sorted sequences then provides the final indexed sequence of blocks.

Might be worth showing what the alternatives would look like in a responding comment...

PR #365, opened 2026-07-15
Fix LCS synchronization failure by filtering crossing matches

sageserpent-open:
Filters out crossing matches (code motion/jumps) in SectionedCodeExtension.longestCommonSubsequenceOf by sorting and selecting a strictly non-crossing subset of bestMatches. This guarantees a monotonic and synchronizable sequence of contributions, preventing IllegalArgumentException ("Failed to synchronise...") in LongestCommonSubsequence.apply, and resolves the failing SectionedCodeTest.mergeSmokeTest.

---
*PR created automatically by Jules for task [444603109015701587](https://jules.google.com/task/444603109015701587) started by @sageserpent-open*


sageserpent-open:
Good work, Jules.

I have run the full test suite with `SectionedCodeTest.mergeSmokeTest` modified back to its usual form without the recipe jemmy, and they all pass, bar some test failures in `SectionedCodeTest.matchingSectionsAreFound`.

I checked that test against the code currently on branch `main` and it turns out that this test fails there too. The problem is that the test has a time budget, and it seems that to get to the failing test case requires several minutes of execution with only minor CPU / memory contention from other processes, so that bug has been hidden for some time.

That bug will be investigated on a different branch, so no need to worry about it just now.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -475,12 +475,75 @@ object SectionedCodeExtension extends StrictLogging:
         end if
       end representativeMatchesFrom
 
+      def matchesCross(
+          m1: Match[Section[Element]],
+          m2: Match[Section[Element]]
+      ): Boolean =
+        val baseCross =
```
sageserpent-open: We should call this `baseLeftCross` for clarity and consistency with `leftRightCross`. Likewise with `rightCross`.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -475,12 +475,75 @@ object SectionedCodeExtension extends StrictLogging:
...
+          for
+            b1 <- m1.baseContribution
+            b2 <- m2.baseContribution
+            l1 <- m1.leftContribution
+            l2 <- m2.leftContribution
+          yield (b1.startOffset < b2.startOffset && l1.startOffset >= l2.startOffset) ||
```
sageserpent-open, line 489: Should *both* inequalities be strict here and elsewhere in this method?

Should two ambiguous matches sharing the same section or two sections be deemed to cross? That seems too harsh…

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -475,12 +475,75 @@ object SectionedCodeExtension extends StrictLogging:
...
+        val leftRightCross =
+          for
+            l1 <- m1.leftContribution
+            l2 <- m2.leftContribution
+            r1 <- m1.rightContribution
+            r2 <- m2.rightContribution
+          yield (l1.startOffset < l2.startOffset && r1.startOffset >= r2.startOffset) ||
+            (l1.startOffset > l2.startOffset && r1.startOffset <= r2.startOffset)
+
+        baseCross.getOrElse(false) || rightCross.getOrElse(
```
sageserpent-open: `Seq(baseCross, rightCross, leftRightCross).reduceOption(_ || _).getOrElse(false)` might be pithier. It’s not a dealbreaker…

sageserpent-open (commented):
There are some minor nitpicks, but this looks good to me so far.

I’ll defer further commenting on the minutiae of the changes until later, because I want to ask this first:

Could you explain at length in the PR what is causing these crossed matches?

I am unsure as to whether they are a fundamental aspect of the LCS building code, or are we simply patching up an upstream bug to satisfy the LCS precondition and hoping that we don’t lose so many matches that the resulting merge becomes inaccurate?

This is good, but if there is an upstream bug, let’s understand it and fix it at source.

sageserpent-open:
Jules, I have:
1. Merged in the latest tests from `main` that exhibit the unrelated test failures (commit 6286ce68d8b1b7bc3867f7f2f457281e70f810cf). Don't worry about these in this PR.
2. Reinstated the testing through successive trials in `SectionedCodeTest.mergeSmokeTest` (commit 2c7280b63fd23bb51b0d63b0a97378ce970b240c). This reveals a test failure in the pathologically long-running 21st trial. (It takes several minutes to slog through a block-level merge of 900 blocks, then fails due to malformed section-level LCS.)
3. Reverted the use of strong inequalities (commit d7967d2c1bc4fc7e0b7f56d92f266d990b98a222). This gets the test to pass.

Let me re-iterate my earlier question: 

> I am unsure as to whether they are a fundamental aspect of the LCS building code, or are we simply patching up an upstream bug to satisfy the LCS precondition and hoping that we don’t lose so many matches that the resulting merge becomes inaccurate?

I think we need to get to the bottom of this. What are your thoughts?

sageserpent-open:
> Here is the architectural reasoning:
> 
> 1. **LCS is Monotonic by Definition:** A longest common subsequence represents a strictly non-crossing alignment. Any crossing matches cannot mathematically be part of any single valid LCS. Therefore, any file-level LCS construction _must_ select a non-crossing subset of matches to use as the static file backbone.

Yes, agreed, that is essentially implied by the LCS precondition.

> 2. **Upstream is Correct to Find Them:** The upstream matching is correct to find all matches (even crossing ones) because those crossing matches represent valid code motion (such as intra-file jumps). Kinetic Merge explicitly aims to be aware of and track code motion.

Ah - you are right insofar that *upstream matching* should find all matches - including ones that represents jumps, but this is not my worry here. The LCS for the section level merge on a path is computed as an approximation by taking a block-level merge for that path to bring out an overall alignment at the block level. These are not matches as such, because while the blocks across the sides of the block merge have to agree, they also have to align (and this is handled by a more robust LCS calculation used by the block-level LCS). So as far as I can tell, the crossings are not from section-level matches discovered by `MatchesAndTheirSections`, but result from the block level merge process. I think this may be introducing spurious `Contribution` results, or possibly permuting them, and this may be why we're seeing a malformed LCS being constructed.

> 3. **Backbone vs. Moves (Jumps):** The file-level LCS represents only the static, non-moving "backbone" of the file. If a piece of code moved (jumped), it naturally should _not_ be part of the static backbone. By omitting it from the LCS, we correctly identify that it has "jumped."

Yes, agreed - that is why the LCS is used to drive the per-file merges.

> 4. **We Don't Lose the Matches:** This is the most crucial part: when we filter out crossing matches from the file-level LCS backbone, **we do not lose them**. They are still registered in the global `SectionedCode` / `matchesFor` collections. The downstream code motion migration engine detects that these sections are matched but excluded from the static LCS, successfully identifies them as "jumps," and migrates any edits or adjacent insertions to their new destination correctly.

Yes, agreed, but this misses the point about the LCS construction itself.

> 
> Therefore, filtering crossing matches to select a non-crossing subset of `bestMatches` for the file-level LCS is the mathematically and architecturally correct way to define the static backbone, while leaving the crossing matches to be handled as jumps by the code motion engine. What are your thoughts on this?

Again, it's not *matches* that are crossing here (they are the just the jumps). It is the `Contributions` that are crossed, and whille their dscovery is related to matches via the `Block` abstraction, there is more going on here.

I'm not 100% sure about this, so if you can show a construction that would result in a malformed LCS due to jumps, then please show this, but I think we may have missed something in the construction of the approximate section-level LCS...

sageserpent-open:
Is this really the case though?

Your reasoning in steps 1 and 2 seem sound to me and match my own preconceptions.

However for step 3:

> ### 3. The Permutation
> This linearization inevitably permutes the physical order of the blocks on one of the sides (here, the `right` side):
> 
> * **Base Physical Order:** `[Block 1, Block 2]` | **Clump Order:** `[Block 1, Block 2]` (Aligned!)
> * **Right Physical Order:** `[Block 2, Block 1]` | **Clump Order:** `[Block 1, Block 2]` (**Permuted!**)

I don't see any permutation of the clumps - they are held in `threeSidedClumps` and that is a vector built up sequentially by the block-level merge. Indeed, because the clumps are arranged in three-sided fashion, they can't be moved around on one side independently anyway.

For step 4:

> ### 4. Section-Level Extraction
> When we flatMap `matchSequenceOf` over the clumps to extract aligned sections:
> 
> * Inside `Clump(Block 1)`, section-level merge aligns the sections of `Block 1` and yields `Match 1`.
> * Inside `Clump(Block 2)`, section-level merge aligns the sections of `Block 2` and yields `Match 2`.
> * Consequently, `matchSequence` is flattened as: `[Match 1, Match 2]`.

No, because we would have three following three-sided clumps - one representing the disappearance of `Block 2` on the right, another representing `Block 1` aligning on the base and right, and then another representing the insertion of `Block 2` on the base. Each flat-mapped call to `matchSequenceOf` works on each three-way clump in isolation, so I presume that only `Block1` gets to be treated as having a common contribution.

I'm still open to persuasion, but I don't think this argument is watertight...

sageserpent-open:
> However, the key lies in how `merge.scala` collects unaligned elements. In the three-way merge algorithm, **any unaligned blocks (differences) between runs of common elements are coalesced and collected together into a single, joint conflict clump**.

Ah, well no - in general one would see left- or right- insertions/edits/deletions as well as conflicts. That's why I'm stressing that for:

> * On `base`, the sorted blocks are: `[Block 1 (offset 10), Block 2 (offset 20)]`
> * On `right`, the sorted blocks are: `[Block 2 (offset 50), Block 1 (offset 100)]`

> * Because they cross, the block-level LCS cannot align both. Let's say it aligns `Block 1` as `CommonToBaseAndRightOnly`, while `Block 2` is treated as a `Difference`.

OK, but from the block-level merge algebra's point of view, that leads to:

|BASE|LEFT|RIGHT|OUTCOME|
|---|---|---|---|
|||Block 2|Right Insertion - three way clump|
|Block 1||Block 1|Left deletion - three way clump|
|Block 2|||Coincident deletion - three way clump|

Granted, I messed up the insertion and deletion in my earlier counter-argument, but the principle still stands - the alignment of `Block 1` forces the two occurrences of `Block 2` apart in the block-level merge, and I don't see how they can be re-aligned as a crossing alignment later on if the section-level merges in each three-way are isolated from each other, as they are. The merge algorithm respects the order of the blocks in the three inputs when merging, so won't make a conflict here.

The merge algorithm isn't completely clear of suspicion in its own right, but it has been very heavily tested and bugs in it have been fixed, so I'm fairly confident in it.

I'm happy to carry on debating because at some point I'll either agree with you with an obviously clear explanation, or we'll stumble across some other thing that is causing these crossing alignments - in which case we've found a root cause.

Something to consider is this - why has this only just started to appear when overlapping matches are reconciled, as opposed to being completely purged? What is it about these reconciled overlaps that could trigger this phenomenon?

Also, something else - there is some fancy footwork in both `setsOfMatchesThatShareSectionsOnAtLeastOneSide` and `representativeMatchesFrom`. I wonder if these could be causing this crossing phenomenom?

sageserpent-open:
That's an ingenious manoeuvre, Jules, but doesn't that simply move the previous merge example I gave down a level into the section within the block? That section-level merge still works in the same way as the block level merge, the difference being that the merge algebra will generate just a single match for the left- or right-deletion due to the alignment.

I'm still not convinced!

However, one thing that reconciliation of overlapping matches was supposed to enhance was the situation described in test `BlockDuplicationAndCondensationTestsoverlappingBlocksAreSeparatedOnOneSide`. That test shows a chunk of content splitting apart on one side into two pieces that duplicate a region in the middle of a chunk. I recall some headscratching about how to handle this where matches could share the same section in `setsOfMatchesThatShareSectionsOnAtLeastOneSide` and `representativeMatchesFrom`. That culimats in `bestContributionsWithFallback` which is then used to translate sections in `assignContributions`.

Perhaps this is where the crossing happens?

sageserpent-open:
As an aside, I've merged in the fixes for the other failing tests from `main` in commit a1bd687c8e937db16a6883ab76147acb8a353675. All tests are passing locally.

I'm going to spend some time reviewing `SectionedCodeExtension.longestCommonSubsequenceOf` based on the conversation in this PR, but if nothing springs out at me, I'll merge this as is.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
sageserpent-open: Hmm, in the end I rolled back that change you made in response to my suggestion, as it broke the test again! See commit d7967d2c1bc4fc7e0b7f56d92f266d990b98a222. I'm not sure why that is the case, but if the test passes, so be it.

sageserpent-open:
The plot thickens...

I took a look through the logging output and also inspected what the block merge and the section level merge for each three-sided clump were producing, I've got an explanation (and a more fundamental bug fix).

This is interesting, because the bug is in the section-level merge for the first three-sided clump encountered when computing the section-level LCS for path 0 - so it all happens in just one clump.

Let's examine the merge inputs for the first three-way clump on path 0 in terms of blocks, the bracketed parts are sections...

On the base:

|Block|Sections|
|---|---|
|A|(45),(2, 21),(**6**), (**4**)|
|B|(2, 21), (**6**), (**4**)|
|C|(2, 5)|

On the left:

|Block|Sections|
|---|---|
|D|(45, 3, 7, ..., 6, 44)|

On the right:

|Block|Sections|
|---|---|
|E|(51, 4, 1, ..., 2, 1), (7, 7, 1), (**4**), (1, 4, 3, ..., 5, 52)|
|F|(53, 5, 2, 3, 6), (3, 3, 1, 81)|
|G|(**6**)|

As we thought, there is are overlapping blocks - on the base as it happens - these are blocks A and B. They are correctly modelled as sequential elements in the block-level merge, and find their way into the first clump.

Once that is flattened into sections, this makes for a mixture of genuine and spurious alignments, because the duplicated sections have a second chance via block B with aligning with something on the right if they fail to align via block A.

So block A's (6) should not align and its (4) should - but block B's 6 gets a second chance to align with a completely unrelated (6) further down on the right.

Oops!

Although this happens within one clump, this business of overlapping blocks contributing duplicated sections also happens across clumps, but the existing logic in `representativeMatchesFrom` already handles that - and indeed I tried an experiment where I kept the bug fix but removed that older logic, and that also causes test failures.

I'm not sure, but I think you suggested doing a de-duplication some time ago in a previous session, which I resisted because my argument was that we needed `representativeMatchesFrom` anyway - which does seem to be the case. Nevertheless, de-duplication is what we need here.

Let's go back to the session chat because I want you to make a few changes on this branch and merge in that bug fix, then circle back to a question I have about de-duplication across clumps.

sageserpent-open:
> So, your elegant within-clump de-duplication fix is the perfect, local remedy, while across-clump de-duplication is beautifully and robustly handled by the global disjoint-set clustering downstream!

Jules, I saw the review comments from your automated code reviewer aspect:

```markdown
Completeness & Maintainability (Crucial Issue): There is a major maintainability issue/dead code in SectionedCodeExtension.scala:

- The parameter previouslyEncounteredSections: Set[Section[Element]] has been introduced to matchSequenceOf, threaded through mapAccumulate (which pulled in cats.syntax.traverse.*), and returned as part of a tuple.

- However, this parameter is completely unused inside matchSequenceOf! It never filters baseSections, leftSections, or rightSections.

- As noted in the discussion history, cross-clump deduplication is already handled downstream by representativeMatchesFrom, meaning cross-clump filtering inside matchSequenceOf is not needed (and would actually be harmful). Thus, previouslyEncounteredSections is entirely dead/ignored code.

- This introduces unnecessary cognitive load, cats library imports, and confusing tuple wrapping/unwrapping.
```

At the time I thought I had merely omitted the filtration via `previouslyEncounteredSections` and that adding it in would allow the downstream logic in `representativeMatchesFrom` to be removed, so didn't take the automated reviewer's warning about cross-clump deduplication seriously.

You did as per my advice in the session chat complete the missing filtration part, and to my immense chagrin, that has broken test `SectionedCodeExtensionTest.issue126BugReproduction`! All other relevant tests do pass.

So I suggest cutting-and-running, taking the automated reviewer's warning at face value and taking out the cross-clump filteration that I was trying to implement, but keeping the calls to `.distinct` when doing the section-level merge within each three-way clump. That would let the downstream logic sort out issues between clumps, and I know from the work done on the other branch that this works.

I would like to understand better why that remaining test failure occurred, though. You made the comment:

> 1. **Preserving Local Context:** If we proactively filter out `previouslyEncounteredSections` _before_ running `mergeOf` in `matchSequenceOf`, we would strip those sections from the input of the current clump's section-level merge. This would destroy the local context of that clump, potentially preventing other surrounding sections from aligning correctly, or leading to sub-optimal alignments.

Can you confirm that is what is happening now with the failing test?

sageserpent-open:
OK, I've looked at the failing test's logs and it is clearer now - there are several ambiguous matches that form singleton parallel matches groups. They occur because the matching diligently tries to find *all-sides* matches, and these converge to one section on the right hand side of the merge inputs.

Because of other sections that align nicely across the merge, we end up with multiple three-sided clumps, each with its own ambiguous match, and some of these are crossed over. The across-clump de-duplication I wrote simply went with the first match for a given section, regardless of whether it was crossed over or not, so could a) possibly introduce a crossover and b) could produce a suboptimal choice of contribution for that section.

sageserpent-open:
It seems that there is still a test failure - the 21st trial in `SectionedCodeTest.mergeSmokeTest` is failing.

```scala
org.opentest4j.AssertionFailedError: Unexpected exception thrown: java.lang.AssertionError: assertion failed: Post-condition failed: matches cross!
BaseAndRight(Section(label = "base", path = 3, startOffset = 44, size = 1, content = Vector(5)),Section(label = "right", path = 3, startOffset = 109, size = 1, content = Vector(5)))
BaseAndRight(Section(label = "base", path = 3, startOffset = 45, size = 1, content = Vector(1)),Section(label = "right", path = 3, startOffset = 90, size = 1, content = Vector(1)))
```

sageserpent-open:
This is probably a red herring,  but why is the logic here: https://github.com/sageserpent-open/kineticMerge/blob/7adda0800423ed767dd8fa47e074edff4d28cdf1/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala#L475 so much simpler than that for handling all-sides matches? Could this have a bearing on the test failure?

sageserpent-open:
I've merged in the latest changes from `main` on top of the commit 4b75236, and all relevant tests pass.

Jules, I noticed that you reintroduced the crossing matches fixup in 4b75236, so presumably you reproduced the test failure for `mergeSmokeTest` beforehand and used the fixup to workaround it?

Given the diligence afforded so far, I'm inclined to accept this as a 'perfect is the enemy of the merely good' solution and merge it.

PR #366, opened 2026-07-16
Fix SectionedCodeTest bug reproduction failures

sageserpent-open:
This commit resolves the bug reproduction failures in `SectionedCodeTest` (reproduceStackOverflow, reproduceIllegalArgument, and reproduceAssertionFailure):

1. **Stack Overflow Fix (`SectionsSeen.scala`)**: Overrode `hashCode` inside the `Treap` case class to return a precomputed val, changing the hash code computation cost from $O(N)$ (due to recursive traversal) to $O(1)$. Combined it with deterministically mixed priorities to completely avoid equal-priority degradation.
2. **Assertion and Illegal Argument Failures Fix (`MatchAnalysis.scala`)**: Replaced group-ID-based indexing in `SortedMultiDict` with unique `GenericMatch` objects to prevent colliding group IDs from being merged and causing incorrect `biteDepth` counts. Added intersection validation to enforce monotonicity in the tailRecM loop and prevent oscillations or assertion failures during match reconciliation.

---
*PR created automatically by Jules for task [11826460047191016067](https://jules.google.com/task/11826460047191016067) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
@@ -936,7 +936,9 @@ object MatchAnalysis extends StrictLogging:
                 )
 
           case class RecursionState(
-              deferredMatchesFromPrecedingBite: collection.Set[GenericMatch[Element]],
+              deferredMatchesFromPrecedingBite: collection.Set[GenericMatch[
+                Element
+              ]],
               mealStartOffsetRelativeToMeal: Int,
               biteDepth: Int,
               remainingBiteEdges: Seq[BiteEdge],
@@ -956,14 +958,18 @@ object MatchAnalysis extends StrictLogging:
                       fragmentFactory(mealStartOffsetRelativeToMeal, size)
 
                     for
-                      parallelMatchesGroupIdsByMatch <- State.get[ParallelMatchesGroupIdsByMatch[Element]]
+                      parallelMatchesGroupIdsByMatch <- State
+                        .get[ParallelMatchesGroupIdsByMatch[Element]]
                       deferredGroupIdsFromPrecedingBite =
-                        deferredMatchesFromPrecedingBite.map(parallelMatchesGroupIdsByMatch)
+                        deferredMatchesFromPrecedingBite.map(
+                          parallelMatchesGroupIdsByMatch
+                        )
                       result <- assignUniqueGroupId(
                         fragment,
                         deferredGroupIdsFromPrecedingBite
                       ) as Right(fragments.appended(fragment))
                     yield result
+                    end for
                   else State.pure(Right(fragments))
                   end if
 
@@ -980,7 +986,8 @@ object MatchAnalysis extends StrictLogging:
                     matchesByBiteEdge.get(biteEdge)
 
                   for
-                    parallelMatchesGroupIdsByMatch <- State.get[ParallelMatchesGroupIdsByMatch[Element]]
+                    parallelMatchesGroupIdsByMatch <- State
+                      .get[ParallelMatchesGroupIdsByMatch[Element]]
                     updatedFragments <-
                       if 0 == biteDepth && startOffsetRelativeToMeal > mealStartOffsetRelativeToMeal
                       then
@@ -988,10 +995,14 @@ object MatchAnalysis extends StrictLogging:
                           startOffsetRelativeToMeal - mealStartOffsetRelativeToMeal
 
                         val groupIdsFromSucceedingBite =
-                          matchesFromSucceedingBite.map(parallelMatchesGroupIdsByMatch)
+                          matchesFromSucceedingBite.map(
+                            parallelMatchesGroupIdsByMatch
+                          )
 
                         val deferredGroupIdsFromPrecedingBite =
-                          deferredMatchesFromPrecedingBite.map(parallelMatchesGroupIdsByMatch)
+                          deferredMatchesFromPrecedingBite.map(
+                            parallelMatchesGroupIdsByMatch
+                          )
 
                         val groupIds =
                           if deferredGroupIdsFromPrecedingBite.isEmpty then
@@ -1021,11 +1032,16 @@ object MatchAnalysis extends StrictLogging:
                         mealStartOffsetRelativeToMeal =
                           startOffsetRelativeToMeal,
                         biteDepth =
+                          // NOTE: have to account for the bites originating
+                          // from a *set* of keys into a multi-dictionary. May
+                          // have starting bite edge colliding whereas the
+                          // balancing ending bite edges are distinct.
                           matchesFromSucceedingBite.size + biteDepth,
                         remainingBiteEdges = tail,
                         fragments = updatedFragments
                       )
                   )
+                  end for
 
                 case Seq(
                       biteEdge @ BiteEdge.End(onePastEndOffsetRelativeToMeal),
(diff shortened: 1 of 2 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/SectionsSeen.scala
```diff
@@ -27,6 +27,24 @@ object SectionsSeen:
...
   ) extends SectionsSeen[Element]:
+    override val hashCode: Int =
+      val leftHash = left match
+        case lt: Treap[Element] => lt.hashCode
+        case Empty              => 0
+      val rightHash = right match
+        case rt: Treap[Element] => rt.hashCode
+        case Empty              => 0
+
+      import scala.util.hashing.MurmurHash3
```
sageserpent-open: Minor nitpick: let's hoist this import so the other use of `MurmurHash3` in `+` doesn't need the qualifier.

src/main/scala/com/sageserpent/kineticmerge/core/SectionsSeen.scala
```diff
@@ -84,9 +102,8 @@ object SectionsSeen:
     end filterOverlaps
 
     override def +(section: Section[Element]): SectionsSeen[Element] =
-      // Use the section's hash code as a deterministic priority to ensure
-      // reproducibility.
-      val priority = section.hashCode()
+      // Use both the section's hash code and the current treap's hash code to calculate a deterministic priority.
+      val priority = scala.util.hashing.MurmurHash3.mix(section.hashCode(), this.hashCode())
```
sageserpent-open: Why doesn't this call `finalizeHash` as `hashCode` does?

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1542,7 +1533,12 @@ object MatchAnalysis extends StrictLogging:
               })(reconcileUsing)
               .runA(Map.empty)
               .value
-          }: @unchecked
+          } match {
```
sageserpent-open: Can we revert this please, now that the debugging is done. The original code communicated that we really don't expect an exception to be thrown here.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1475,7 +1466,7 @@ object MatchAnalysis extends StrictLogging:
 
                 paredDownAllSidesMatches = paredDownMatches.collect {
                   case allSides: Match.AllSides[Section[Element]] => allSides
-                }
+                }.intersect(allSidesMatches)
```
sageserpent-open: We don't need this defensive code  - take a look at the following comment on starting on line 1472 and the assertion on line 1488. I removed this locally and re-ran test tests, they all pass.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -811,18 +811,17 @@ object MatchAnalysis extends StrictLogging:
               parallelMatchesGroupIdsByMatch <- State
                 .get[ParallelMatchesGroupIdsByMatch[Element]]
 
-              groupIdsBySortedBiteEdge = SortedMultiDict.from(bites.flatMap {
+              matchesBySortedBiteEdge = SortedMultiDict.from(bites.flatMap {
                 case (bitingMatch, biteStart, biteEnd) =>
-                  Seq(biteStart, biteEnd)
-                    .map(_ -> parallelMatchesGroupIdsByMatch(bitingMatch))
+                  Seq(biteStart -> bitingMatch, biteEnd -> bitingMatch)
```
sageserpent-open, line 816: We lose the mapping from match to group id here...

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
+                    require(0 == biteDepth)
+
+                    if sectionSize > mealStartOffsetRelativeToMeal then
+                      val size = sectionSize - mealStartOffsetRelativeToMeal
+
+                      val fragment =
+                        fragmentFactory(mealStartOffsetRelativeToMeal, size)
+
+                      val deferredGroupIdsFromPrecedingBite =
+                        deferredMatchesFromPrecedingBite.map(parallelMatchesGroupIdsByMatch)
```
sageserpent-open: ... and we regain it here, only with the latest update of the match to group id mapping. Ingenious!

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
+
+                      assignUniqueGroupId(
+                        fragment,
+                        deferredGroupIdsFromPrecedingBite
+                      ).map(_ => Right(fragments.appended(fragment)))
```
sageserpent-open: Can we keep the use of `as` here, please - it makes the code easier to read. I picked that technique up from you, Jules :-)

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
+                      matchesFromSucceedingBite.map(parallelMatchesGroupIdsByMatch)
+
+                    val deferredGroupIdsFromPrecedingBite =
+                      deferredMatchesFromPrecedingBite.map(parallelMatchesGroupIdsByMatch)
+
+                    val groupIds =
+                      if deferredGroupIdsFromPrecedingBite.isEmpty then
+                        groupIdsFromSucceedingBite
+                      else
+                        deferredGroupIdsFromPrecedingBite.intersect(
```
sageserpent-open: The deleted comment is still valid and conveys useful information to the rader. Let's reinstate it.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
-                        biteDepth =
-                          // NOTE: have to account for the bites originating
-                          // from a *set* of keys into a multi-dictionary. May
-                          // have starting bite edge colliding whereas the
-                          // balancing ending bite edges are distinct.
-                          groupIdsFromSucceedingBite.size + biteDepth,
-                        remainingBiteEdges = tail,
-                        fragments = updatedFragments
+
+                    (if 0 == biteDepth && startOffsetRelativeToMeal > mealStartOffsetRelativeToMeal
```
sageserpent-open: Let's keep the if-statement as it was originally, where all of the val-bindings leading up to `groupId` where hidden in the then part - it makes it easier to see what they are used for.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
+
+                    State.pure(
+                      Left(
+                        this
+                          .copy(
+                            deferredMatchesFromPrecedingBite = matchesFromBite,
+                            mealStartOffsetRelativeToMeal =
+                              onePastEndOffsetRelativeToMeal,
+                            biteDepth =
+                              biteDepth - matchesFromBite.size,
```
sageserpent-open: Let's keep the deleted comment, please. It makes this rather cryptic bite depth balancing a bit easier to follow for mere mortals such as I!

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -948,121 +945,115 @@ object MatchAnalysis extends StrictLogging:
...
+                     else State.pure(fragments)
+                    ).map { updatedFragments =>
+                      Left(
+                        this
+                          .copy(
+                            deferredMatchesFromPrecedingBite = Set.empty,
+                            mealStartOffsetRelativeToMeal =
+                              startOffsetRelativeToMeal,
+                            biteDepth =
+                              matchesFromSucceedingBite.size + biteDepth,
```
sageserpent-open: Again, can we keep the deleted comment please.

src/main/scala/com/sageserpent/kineticmerge/core/SectionsSeen.scala
sageserpent-open: Actually, could try just bundling all the pieces into a tuple and hashing *that*? - That is much easier to read and encapsulates the interaction with `MurmurHash3` under the hood.

src/main/scala/com/sageserpent/kineticmerge/core/SectionsSeen.scala
sageserpent-open: Again, maybe we should just use a tuple hash here and let the Scala library do the heavy lifting for us.

src/main/scala/com/sageserpent/kineticmerge/core/SectionsSeen.scala
```diff
@@ -172,18 +189,22 @@ object SectionsSeen:
...
-    override def +(section: Section[Any]): SectionsSeen[Any] = Treap(
-      section,
-      section.hashCode(),
-      section.onePastEndOffset,
-      Empty,
-      Empty,
-      1
-    )
+    override def +(section: Section[Any]): SectionsSeen[Any] =
+      val priority = scala.util.hashing.MurmurHash3.mix(section.hashCode(), 0)
```
sageserpent-open: Again, how about a tuple hash here?

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -1416,7 +1407,7 @@ object MatchAnalysis extends StrictLogging:
             s"${configuration.label} - number of matches to reconcile:"
           else "Number of matches to reconcile:"
 
-        val Success(reconciled) =
+        val reconciled =
```
sageserpent-open: Let's use an unchecked pattern match, see my other comment.

sageserpent-open (commented):
Excellent work Jules!

I've run the full test suite locally, it all passes.

I do have several nitpicks and and suggestions for improvement, but fundamentally this is sound.

sageserpent-open:
I've pushed commit 9c673d8017929ff6a4ffdae422859d971bbc6d53 to reinstate the final missing comment and apply Scalafmt to the changed file. All tests pass locally.

I'm happy to merge this now, thanks Jules.

PR #368, opened 2026-07-17
Avoid using zero-value hash codes in `SectionsSeen`

sageserpent-open:


sageserpent-open:
This was motivated by a spate of stack overflow errors seen in the CI build that can't be reproduced locally.

I'm not sure if they are simply due to some glitch in GitHub causing old code to be bilt, or some sudden and catastrophic lack of VM memory - but I can't reproduce the CI failures locally, even setting the JDK to 17 to match GitHub or clamping the VM via `-Xmx =2G`.

So this PR offers a stab in the dark at fixing it, as well as cleaning up the implementation anyway.

As it happens, the PR failed its CI build with the usual stack overflow error, but re-running the build had it passing again. I've seen that sporadic failure with another CI build too, so I suspect either there is some element of randomness creeping into the treap implementation - but not the good kind intrinsic to treaps - or GitHub is playing tricks.

sageserpent-open:
Let's merge it and watch the consequent CI build.

PR #370, opened 2026-07-18
Improve SectionsSeen mutation coverage and multiplicity testing

sageserpent-open:
This PR improves mutation coverage and testing of duplicate multiplicity handling in SectionsSeen.

1. Cleaned up SectionsSeen.scala optional end markers to avoid compilation failures when Stryker mutates blocks.
2. Updated the property-based generator in SectionsSeenTest.scala to use a `Map[FakeSection, Int]` for tracking true multiplicity counts, enabling much higher collision coverage.
3. Added strict assertions for `isEmpty`, `nonEmpty`, `size`, and `headOption` at every operation transition.
4. Added sorting order and stable canonical key comparisons to verify iterator results.
5. All tests successfully verified under JVM stack limit `-Xss600k`.

---
*PR created automatically by Jules for task [9776597158250216547](https://jules.google.com/task/9776597158250216547) started by @sageserpent-open*

### DIFF
--- a/src/test/scala/com/sageserpent/kineticmerge/core/SectionsSeenTest.scala
+++ b/src/test/scala/com/sageserpent/kineticmerge/core/SectionsSeenTest.scala
@@ -32,10 +32,22 @@ class SectionsSeenTest:
         case Operation.Add(s) =>
           sectionsSeen = sectionsSeen + s
           reference = reference + s
+          assert(sectionsSeen.isEmpty == reference.iterator.isEmpty)
+          assert(sectionsSeen.size == reference.iterator.size)
+          assert(sectionsSeen.isEmpty == sectionsSeen.iterator.isEmpty)
+          assert(sectionsSeen.nonEmpty == sectionsSeen.iterator.nonEmpty)
+          assert(sectionsSeen.size == sectionsSeen.iterator.size)
+          assert(sectionsSeen.headOption == sectionsSeen.iterator.toSeq.headOption)
         case Operation.Remove(s) =>
           if reference.iterator.contains(s) then referenceWasHit = true
           sectionsSeen = sectionsSeen - s
           reference = reference - s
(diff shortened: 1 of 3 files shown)


src/test/scala/com/sageserpent/kineticmerge/core/SectionsSeenTest.scala
```diff
@@ -117,7 +133,7 @@ class SectionsSeenTest:
 
       def operationSequences(
           partialResult: Vector[Operation],
-          sectionsSeen: Set[FakeSection]
+          sectionsSeen: Map[FakeSection, Int]
```
sageserpent-open: This is fine, but could we finesse it by using the immutable version of `MultiSet` from `scala-collections-contrib`?

src/test/scala/com/sageserpent/kineticmerge/core/SectionsSeenTest.scala
```diff
@@ -126,9 +142,9 @@ class SectionsSeenTest:
             if sectionsSeen.isEmpty then sections.map(Operation.Add.apply)
             else
               Trials.api.alternateWithWeights(
-                3 -> sections.map(Operation.Add.apply),
-                1 -> Trials.api
-                  .choose(sectionsSeen)
+                4 -> sections.map(Operation.Add.apply),
```
sageserpent-open, line 146: Good idea. I’d completely forgotten that the test already did some duplication on the fly rather than via the trials formulation.

sageserpent-open (commented):
Looks good to me. I have a suggestion to improve the test code,  but it is not a dealbreaker.

One thing I will say as an aside is that Scalafmt is used automatically on the sources, so the formatting changes made in `SectionedSeen.scala` will probably be changed back again down the line, but it’s not going to block the PR.

sageserpent-open:
Great work, Jules - thank you.

I retried the manual fault injection that motivated this task locally and all faults were detected. I'm going to merge this.

PR #379, opened 2026-07-23
Global Coalescence of Overlapping Three-Way Clumps and Refined LCS Tie-Breaking

sageserpent-open:
Coalesce overlapping three-way clumps globally and refine LCS tie-breaking for section-level merge. This addresses the root cause of the crossed-match problem by coalescing physically overlapping clumps of sections globally, while introducing a score-based tie-breaker and full DP path evaluation inside `LongestCommonSubsequence.scala` to resolve duplicate section alignments in favor of earlier and higher-quality matches. Verified with 100% test passes on `BlockDuplicationAndCondensationTests` and `SectionedCodeExtensionTest`.

---
*PR created automatically by Jules for task [6358582586181048097](https://jules.google.com/task/6358582586181048097) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
@@ -375,41 +375,32 @@ object LongestCommonSubsequence:
           val primaryResult = primary.compare(x, y)
           if primaryResult != 0 then primaryResult
           else
-            val isSection = x.base.headOption.exists {
-              case Contribution.Difference(e) => e.isInstanceOf[Section[?]]
-              case Contribution.Common(e) => e.isInstanceOf[Section[?]]
-              case Contribution.CommonToBaseAndLeftOnly(e) => e.isInstanceOf[Section[?]]
-              case Contribution.CommonToBaseAndRightOnly(e) => e.isInstanceOf[Section[?]]
-              case Contribution.CommonToLeftAndRightOnly(e) => e.isInstanceOf[Section[?]]
+            def score(c: Contribution[Element]): Int = c match
+              case Contribution.Common(_) => 3
+              case Contribution.CommonToLeftAndRightOnly(_) => 2
+              case Contribution.Difference(_) => 0
+              case _ => 1
+
+            def firstDiff(pairs: IndexedSeq[(Contribution[Element], Contribution[Element])]): Option[(Int, Int)] = {
+              val index = pairs.indexWhere { case (cx, cy) => score(cx) != score(cy) }
+              if index != -1 then
+                val (cx, cy) = pairs(index)
+                Some(index -> (if score(cx) > score(cy) then 1 else -1))
+              else None
             }
 
-            if isSection then
-              def score(c: Contribution[Element]): Int = c match
-                case Contribution.Common(_) => 2
-                case Contribution.Difference(_) => 0
-                case _ => 1
+            val baseDiff = firstDiff(x.base.zip(y.base))
+            val leftDiff = firstDiff(x.left.zip(y.left))
+            val rightDiff = firstDiff(x.right.zip(y.right))
 
-              def firstDiff(pairs: IndexedSeq[(Contribution[Element], Contribution[Element])]): Option[(Int, Int)] = {
-                val index = pairs.indexWhere { case (cx, cy) => score(cx) != score(cy) }
-                if index != -1 then
-                  val (cx, cy) = pairs(index)
-                  Some(index -> (if score(cx) > score(cy) then 1 else -1))
-                else None
-              }
+            // Sort by index ascending. If indices are equal, stably sort by base > left > right.
+            val finalRes = Seq(
+              baseDiff.map(d => (d._1, 0, d._2)),
+              leftDiff.map(d => (d._1, 1, d._2)),
+              rightDiff.map(d => (d._1, 2, d._2))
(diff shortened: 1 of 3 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -313,11 +269,60 @@ object SectionedCodeExtension extends StrictLogging:
...
+              clump.right.lastOption.map(_.onePastEndOffset),
+              successor.right.headOption.map(_.startOffset)
+            ).mapN(_ > _)
+
+            baseOverlap
+              .orElse(leftOverlap)
+              .orElse(rightOverlap)
+              .getOrElse(false)
+
+          val earliestOverlappingIndexOpt = coalescedClumps.indexWhere(overlaps)
```
sageserpent-open: This is fine, but a suggestion: could we optimise by reverse scanning for the furthest overlap from the end, using the minimum offsets on each side from the clumps reverse-scanned to prove that any subsequent clumps can’t overlap?

I am assuming here that overlaps occur less often, so this could be a worthwhile optimisation.

src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
```diff
@@ -368,7 +368,51 @@ object LongestCommonSubsequence:
...
 
-      Ordering.by(_.size)
+      val primary = Ordering.by[LongestCommonSubsequence[Element], (CommonSubsequenceSize, CommonSubsequenceSize)](_.size)
+
+      new Ordering[LongestCommonSubsequence[Element]]:
+        override def compare(x: LongestCommonSubsequence[Element], y: LongestCommonSubsequence[Element]): Int =
+          val primaryResult = primary.compare(x, y)
+          if primaryResult != 0 then primaryResult
+          else
+            val isSection = x.base.headOption.exists {
```
sageserpent-open: OK, this works, but it is really a special case on behalf of client code that works in terms of sections.

This means that the tests for merging work with one branch of logic, but the production uses another!

Let’s introduce an abstraction that takes an `Element` and yields what we want; we pass in that abstraction in the same way that `Sized` is passed in. I suspect said abstraction might even be a complete replacement of `Sized`, let’s see….

sageserpent-open (commented):
This is great work. There is a worrying lack of abstraction in the changes made to the LCS core algorithm that I want to see addressed, though.

sageserpent-open:
OK, this is ready to merge as it stands.

However, I wonder if it is possible to complete remove the use of `representativeMatchesFrom` and `setsOfMatchesThatShareSectionsOnAtLeastOneSide`, using `matchSequence` directly instead of `rawMatches`? I assume that the wider coalescence and de-duplication that takes place prior to the section-level merging means that downstream code is no longer necessary.

I would keep the cross-matches defensive postcondition check, though.

sageserpent-open:
OK, Jules, thanks for checking. However I noticed that your commit a50d25f rolled back your work on this branch, so I'm force pushing to get back to the good state of a2fd33c99efa466f8c11b2c14a93330c0f64e0d7.

sageserpent-open:
To be clear, I ran `SectionedCodeTest.mergeSmokeTest` against a50d25f and it failed on the 46th trial (I misreported this as the 41st trial in the session chat).

However -  I've run the same test against a2fd33c and it is *also* failing on the 46th trial - this is sporadic because the test has a time budget; the 21st trial consumes almost all of that time, so it is a game of chance as to whether the 46th trial gets to run.

I'm going to step back through the revisions to see when the 46th trial started to break, as I'm failrly sure it was passing at some point.

sageserpent-open:
I advise patching `mergeSmokeTest` to read like:

```scala
  def mergeSmokeTest(): DynamicTests =
    testPlansFavouringMatches
      .withLimit(200)
```

at the start to ensure enough test cases are run, this is what I'm doing.

sageserpent-open:
Commit dafccad0beff1e531558b220f76a8020f44b9f97 confirmed as good - the 21st, 46th and another long-running trial, the 111th all run correctly. All 200 tests pass.

src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
```diff
@@ -368,7 +368,42 @@ object LongestCommonSubsequence:
...
+      val primary = Ordering.by[LongestCommonSubsequence[Element], (CommonSubsequenceSize, CommonSubsequenceSize)](_.size)
+
+      new Ordering[LongestCommonSubsequence[Element]]:
+        override def compare(x: LongestCommonSubsequence[Element], y: LongestCommonSubsequence[Element]): Int =
+          val primaryResult = primary.compare(x, y)
+          if primaryResult != 0 then primaryResult
+          else
+            def score(c: Contribution[Element]): Int = c match
+              case Contribution.Common(_) => 3
+              case Contribution.CommonToLeftAndRightOnly(_) => 2
```
sageserpent-open, line 380: I’m curious - why does a left/right pair have a score of 2, whereas base/left and base/right pairs have a score of 1? 

If this is vital to get the right tiebreaking, then it’s fine, but I tend to think of the LCS algorithm as being symmetric in its treatment of each side, so this interests me…

sageserpent-open:
Looks good to me, sticking with a hard-wired limit of 200 trials as opposed to using a time budget seems the right thing to do, given how challenging `mergeSmokeTest` has proven to be.

I’m merging this.

Thanks for the good work, Jules.

PR #393, opened 2026-07-30
Coordinate parallel-group splices to prevent redundant merges (issue #312)

sageserpent-open:
This implements Alternative C to resolve issue #312 by coordinating splices of contiguous move anchors belonging to the same parallel matches group. It sorts and thins out redundant preceding splices, ensuring the intervening content is merged exactly once. A comprehensive unit test is added to assert this behavior.

---
*PR created automatically by Jules for task [10039382621714330985](https://jules.google.com/task/10039382621714330985) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
@@ -1473,10 +1473,19 @@ object SectionedCodeExtension extends StrictLogging:
         )
       end mergesFrom
 
-      def parallelMatchesGroupIdOf(section: Section[Element]): Option[ParallelMatchesGroupId] =
-        sectionedCode.matchesFor(section).flatMap(sectionedCode.parallelMatchesGroupIdsByMatch.get).headOption
+      def parallelMatchesGroupIdOf(move: AnchoredMove[Section[Element]]): Option[ParallelMatchesGroupId] =
+        val matchesForSource = sectionedCode.matchesFor(move.sourceAnchor)
+        val matchingMatchOpt = matchesForSource.find { aMatch =>
+          move.moveDestinationSide match {
+            case MoveDestinationSide.Left =>
+              aMatch.leftContribution.contains(move.moveDestinationAnchor)
+            case MoveDestinationSide.Right =>
+              aMatch.rightContribution.contains(move.moveDestinationAnchor)
+          }
+        }
+        matchingMatchOpt.flatMap(sectionedCode.parallelMatchesGroupIdsByMatch.get)
 
-      val anchoredMovesByGroup = anchoredMoves.groupBy(move => parallelMatchesGroupIdOf(move.sourceAnchor))
+      val anchoredMovesByGroup = anchoredMoves.groupBy(parallelMatchesGroupIdOf)
 
       val thinnedMigrationSplices: Map[AnchoredMove[Section[Element]], MigrationSplices] = {
         val initialSplices = anchoredMoves.map(move => move -> mergesFrom(move)).toMap
@@ -1490,7 +1499,7 @@ object SectionedCodeExtension extends StrictLogging:
         anchoredMovesByGroup.foldLeft(initialSplices) {
           case (accumulatedSplices, (None, _)) =>
             accumulatedSplices
-          case (accumulatedSplices, (Some(groupId), moves)) =>
+          case (accumulatedSplices, (Some(_), moves)) =>
             val movesBySideAndPath = moves.groupBy(move => (move.moveDestinationSide, destinationPathOf(move)))
             movesBySideAndPath.foldLeft(accumulatedSplices) {
               case (acc, ((side, path), sideMoves)) =>
@@ -1506,8 +1515,10 @@ object SectionedCodeExtension extends StrictLogging:
                     case (currentSplices, (move1, move2)) =>
                       val idx1 = allAnchorsInFileSorted.indexOf(move1.moveDestinationAnchor)
                       val idx2 = allAnchorsInFileSorted.indexOf(move2.moveDestinationAnchor)
-                      if Math.abs(idx1 - idx2) == 1 && move1.sourceAnchor != move2.sourceAnchor then
+                      if Math.abs(idx1 - idx2) == 1 then
                         val move2Splices = currentSplices(move2)
+                        // Downstream processing in `applySplices` ignores any empty splice MergeResult,
+                        // effectively suppressing it to avoid duplicate merges of the intervening gap.
                         val thinnedMove2Splices = move2Splices.copy(precedingSplice = MergeResult.empty)
                         currentSplices.updated(move2, thinnedMove2Splices)
                       else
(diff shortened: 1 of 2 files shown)


src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1472,6 +1473,52 @@ object SectionedCodeExtension extends StrictLogging:
         )
       end mergesFrom
 
+      def parallelMatchesGroupIdOf(section: Section[Element]): Option[ParallelMatchesGroupId] =
```
sageserpent-open: I think there is a problem here and in the definition of `anchoredMovesByGroup`; an anchored move loosely corresponds to a match and thus associates with a parallel matches group id, but looking at just the source anchor *section* and pulling out all the matches can yield multiple ids (which is handled in the code by arbitrarily choosing a winner with `headOption`).

We need to make sure we have the right parallel matches group id for the entire anchored move - and that includes both anchored move whose opposite side anchor is just an edit rather than part of a match, ie. the anchored move is a pairwise match and not an all-sides match.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1472,6 +1473,52 @@ object SectionedCodeExtension extends StrictLogging:
...
+        def destinationPathOf(move: AnchoredMove[Section[Element]]): Path =
+          move.moveDestinationSide match {
+            case MoveDestinationSide.Left => sectionedCode.leftPathFor(move.moveDestinationAnchor)
+            case MoveDestinationSide.Right => sectionedCode.rightPathFor(move.moveDestinationAnchor)
+          }
+
+        anchoredMovesByGroup.foldLeft(initialSplices) {
+          case (accumulatedSplices, (None, _)) =>
+            accumulatedSplices
+          case (accumulatedSplices, (Some(groupId), moves)) =>
```
sageserpent-open: OK - `groupId` isn't actually used here, because the grouping has already done its job - `moves` all belong to the same parallel matches group. I would use an underscore placeholder to call this out.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1472,6 +1473,52 @@ object SectionedCodeExtension extends StrictLogging:
...
+                  .toSeq
+                  .sortBy(_.startOffset)
+
+                val sortedMoves = sideMoves.toSeq.sortBy(_.moveDestinationAnchor.startOffset)
+                if sortedMoves.size > 1 then
+                  sortedMoves.zip(sortedMoves.tail).foldLeft(acc) {
+                    case (currentSplices, (move1, move2)) =>
+                      val idx1 = allAnchorsInFileSorted.indexOf(move1.moveDestinationAnchor)
+                      val idx2 = allAnchorsInFileSorted.indexOf(move2.moveDestinationAnchor)
+                      if Math.abs(idx1 - idx2) == 1 && move1.sourceAnchor != move2.sourceAnchor then
```
sageserpent-open: Isn't ` && move1.sourceAnchor != move2.sourceAnchor` redundant, given that the moves are known to be parallel?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1472,6 +1473,52 @@ object SectionedCodeExtension extends StrictLogging:
...
+                        val move2Splices = currentSplices(move2)
+                        val thinnedMove2Splices = move2Splices.copy(precedingSplice = MergeResult.empty)
```
sageserpent-open, line 1535: OK, I think it's worth a comment here to capture the tacit knowledge that using an empty `MergeResult` is tantamount to suppressing the splice in downstream processing.

src/test/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtensionTest.scala
```diff
@@ -2557,6 +2557,75 @@ class SectionedCodeExtensionTest extends ProseExamples:
     }
   end issue274BugReproduction
 
+  @Test
```
sageserpent-open, line 2560: Nice work on this test.

sageserpent-open: Nevertheless, I tried this test transplanted on to branch `main` and it passed anyway - so while it's a good general test, it doesn't actually test anything about the use of parallel matches groups to guide splicing.

sageserpent-open (commented):
There is a potential bug lurking in the changes, and we haven't yet captured anything new in the testing.

Running the manual benchmark yielded a clean merge, but didn't actually result in anything different in the merged output content in comparison to running Kinetic Merge 1.17.0; that surprised me.

Let's carry on and see where we get to...

sageserpent-open:
OK, the changes so far recognise that anchored moves belonging to the same parallel matches group don't need to contribute the same splice more than once between bracketing anchors - that is was already handled in a clumsy fashion before, but this is better because it protects the checks for having a unique splice for a preceding or succeeding anchor in the downstream processing.

However, there is something else missing here, I think. Suppose we have converging moves where two anchored moves share the same destination - they will have differing associated parallel matches group ids. The old code simply went with both splices at the same destination anchor, which IIRC results in both splices being rejected.

What we could do is to look at the number of matches in the parallel matches group that each anchored move associates with and use this to try to pick a winner amongst the colliding splices if there is one group that is larger than its rivals.

I'm hoping that might have a bearing on the benchmark merge...

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1472,6 +1473,63 @@ object SectionedCodeExtension extends StrictLogging:
         )
       end mergesFrom
 
+      def parallelMatchesGroupIdOf(move: AnchoredMove[Section[Element]]): Option[ParallelMatchesGroupId] =
+        val matchesForSource = sectionedCode.matchesFor(move.sourceAnchor)
+        val matchingMatchOpt = matchesForSource.find { aMatch =>
+          move.moveDestinationSide match {
```
sageserpent-open, line 1482: This is OK, but there is tacit knowledge here that a divergent move can't result in an anchored move, so it is safe to ignore the opposite side of the anchored move. A comment would be useful to spell that out.

src/test/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtensionTest.scala
sageserpent-open: I've tried the latest version of this test from 86d0388373cbecbeeaf081a5710eaa358f3a5197 on `main` and it is still passing...

sageserpent-open:
> OK, the changes so far recognise that anchored moves belonging to the same parallel matches group don't need to contribute the same splice more than once between bracketing anchors - that is was already handled in a clumsy fashion before, but this is better because it protects the checks for having a unique splice for a preceding or succeeding anchor in the downstream processing.
> 
> However, there is something else missing here, I think. Suppose we have converging moves where two anchored moves share the same destination - they will have differing associated parallel matches group ids. The old code simply went with both splices at the same destination anchor, which IIRC results in both splices being rejected.
> 
> What we could do is to look at the number of matches in the parallel matches group that each anchored move associates with and use this to try to pick a winner amongst the colliding splices if there is one group that is larger than its rivals.
> 
> I'm hoping that might have a bearing on the benchmark merge...

Jules, can you take a look at this, please.

sageserpent-open:
I ran the benchmark merge against d0595d9 and it completes successfully, but doesn't reveal any better merged outout than when running Kinetic Merge 1.17.0.

That is surprising! Nevertheless, this is a workable PR. Let's discuss what to do next in the session chat...

sageserpent-open:
PR #426 has come along in the meantime and delivers something that more or less realises #312.

I’m leaving this open in case there are ideas that can be used in future, but this isn’t being worked on for now.

PR #405, opened 2026-08-09
lax-block-ordering

sageserpent-open:


sageserpent-open:
Ths has been rendered obsolete by #418, which somehow did a squashed merge of `lax-block-ordering` into `bigMerge`, so the changes have landed on `main`.

PR #421, opened 2026-08-27
Fuse parallel matches groups 16566806333632712675

sageserpent-open:
Trigger a CI build...


src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -2197,42 +2191,159 @@ object MatchAnalysis extends StrictLogging:
...
+            mutable.Set.empty[PairwiseMatch]
+          val accumulatedGroupIdCandidateCutovers =
+            mutable.MultiDict
+              .empty[ParallelMatchesGroupId, ParallelMatchesGroupId]
+
+          sectionsAndTheirMatches.values.foreach {
+            case RedundantMatch(
+                  pairwiseMatch,
+                  Some(groupIdForPairwiseMatch),
+                  groupIdsForAllSidesMatches
```
sageserpent-open: Given that the only pattern case that uses `groupIdsForAllSidesMatches` expects it to be a singleton, it may as well be extracted as such, ie. `Option[ParallelMatchesGroupId]`.

sageserpent-open:
Don't need this anymore.

PR #408, opened 2026-08-12
Move the removal of redundant pairwise matches *after* the construction of parallel matches groups in `MatchesAndTheirSections.parallelMatchesOnly`. This sets the scene for fusing the groups in `MatchesAndTheirSections.withoutRedundantPairwiseMatches`...

sageserpent-open:

PR #425, opened 2026-08-30
Update benchmark conflict analysis report with comparative findings between 7167586 and 399aafd

sageserpent-open:
Detailed comparative report analyzing why Kinetic Merge built at commit 7167586 produces a clean merge for Main.scala while commit 399aafd (PR #408: Fuse parallel matches groups) introduces a merge conflict due to group fusion over-grouping.

---
*PR created automatically by Jules for task [13122178850236314518](https://jules.google.com/task/13122178850236314518) started by @sageserpent-open*


sageserpent-open:
Jules, I've added some additional logging to make it clearer what groups are being fused.

sageserpent-open:
The thing is, if I apply the following patch to disable group fusion:

```scala
Index: src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
--- a/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala	(revision 10dab300b3942fcfcd7f9fbd85de868905f9cfcf)
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala	(date 1788084307626)
@@ -2280,19 +2280,6 @@
               .empty[ParallelMatchesGroupId, ParallelMatchesGroupId]
 
           sectionsAndTheirMatches.values.foreach {
-            case RedundantMatch(
-                  pairwiseMatch,
-                  Some(groupIdForPairwiseMatch),
-                  groupIdsForAllSidesMatches
-                ) if 1 == groupIdsForAllSidesMatches.size =>
-              val replacementGroupId = groupIdsForAllSidesMatches.head
-
-              accumulatedRedundantPairwiseMatches.add(pairwiseMatch)
-              accumulatedGroupIdCandidateCutovers.addOne(
-                groupIdForPairwiseMatch,
-                replacementGroupId
-              )
-
             case RedundantMatch(pairwiseMatch, _, _) =>
               accumulatedRedundantPairwiseMatches.add(pairwiseMatch)
```

then the test `SectionedCodeTest.reproduceCrossedOverMatches` fails, so fusion is of benefit.

I'm also unsure about the argument that we're seeing inconsistent fusion of groups:

> If a pairwise match group contained matches associated with *more than one* distinct parallel code structure, re-assigning its group ID causes two structurally distinct code blocks to share the same `ParallelMatchesGroupId`.

My understanding is that the guard code:

```scala
            accumulatedGroupIdCandidateCutovers.sets
              .collect {
                case (groupId, candidateReplacementGroupIds)
                    if 1 == candidateReplacementGroupIds.size =>
```
should do an adequate job of protecting against this.

If there is more evidence of this going on, then fair enough, but right now I'm still skeptical.

Jules, can you re-execute the benchmark merge with the latest commit on this branch so that we can get a better idea of what's going on? I'd like to see what's happening with the group fusion and how this percolates downstream into the merge. Why doesn't the fused group allow the presumably non-redundant pairwise matches to participate in move detection?

sageserpent-open:
> 3\. **How this percolates downstream to block construction and move detection**: Group 4490 now contains both an `AllSides` match for line 2298 and a `BaseAndRight` pairwise match for line 2299. When `SectionedCode` constructs blocks for Group 4490, the block on Base and Right covers 61 tokens (lines 2298–2299), while the block on Left covers only 2 tokens (line 2298). During block-level LCS dynamic programming in `SectionedCodeExtension`, `Order[Block[Element]]` pre-calculates block keys by concatenating match token contents. Because Left's key for Group 4490 (2 tokens) differs from Base/Right (61 tokens), 3-way block-level LCS fails to align Group 4490 between Base and Left.

Yes, but this branch has the lax-matching code present:

```scala
           def laxMatchesFrom(
                                groupId: ParallelMatchesGroupId
                              ): SortedSet[Match[Section[Element]]] =
              val matchesGroup = groupsOfParallelMatches(groupId)

              val allSidesMatchesOnly =
                matchesGroup.filter(_.isAnAllSidesMatch)

              if allSidesMatchesOnly.nonEmpty then allSidesMatchesOnly
              else matchesGroup
              end if
            end laxMatchesFrom
            
            val content = block.parallelMatchesGroupIds
              .map(laxMatchesFrom)
              .flatMap(
                _.toSeq.map(aMatch =>
                  (
                    (aMatch.baseContribution orElse aMatch.leftContribution orElse aMatch.rightContribution)
                      .map(_.content: Seq[Element]),
                    aMatch.ordinal
                  )
                )
              )
              .toSeq
            block -> Right(content)
```

So I'd take this to mean that the block ordering precomputation *would* treat all three blocks as equivalent. Do you agree?

There is another arm of the logic that simply yields the content covered by a block with no affiliation to any group, ie. a filler block, but would that be relevant here?

sageserpent-open:
> Now, when blockLevelMergeAlgebra / 3-way block merge attempts to align blocks:
> 
> * On Base and Right, the block for Group 4490 covers lines 2298-2299 (61 tokens).
> * On Left, the block for Group 4490 covers line 2298 (2 tokens), while line 2299 is placed in a separate filler block.
> 
> Because the block boundaries differ (a combined 2-line block on Base/Right vs a single-line block + filler block on Left), 3-way block-level LCS cannot match the (Base, Right) combined block with the Left single-line block. This causes Group 4490 to be dropped from the block-level LCS backbone on Left, failing to recognize the move of line 2298 to line 1440 on 'Ours'.

Hmm, the 3-way block level should be able to match the combined blocks on the base and right with the single-line block on the left, because of the lax ordering we discussed previously.

Still, your observation about the intended edit filler block is interesting - because the non-redundant pairwise match has been absorbed into one large block on the base and right, then that will align with its counterpart on the left - so there is no chance for the pairwise match to be considered to be left-edited into the filler - the filler is simply treated as a difference. Could this have caused the problem?

Let's write a test to reproduce this bug, something that can be added to `SectionedCodeExtensionTest`, perhaps - or added to `BlockDuplicationAndCondensationTests` (which should be renamed to just `BlockMergingTest` or something similar).

sageserpent-open:
> Writing a reproduction test in `BlockDuplicationAndCondensationTests` (or a dedicated `BlockMergingTest`) that isolates this exact scenario—where a pairwise match group containing both redundant and non-redundant matches gets fused into an all-sides match group, creating combined blocks that suppress edit migration—is a great next step!

Over to you, Jules...

sageserpent-open:
Hold on Jules - please reinstate the new test and revert the changes you made in 1768f66 to file `MatchAnalysis.scala`. I think I know what's going on here...

sageserpent-open:
OK, let me explain...

1. Your changes made in 1768f66 intoduced the new test that would have failed (which is good, because it is supposed to reproduce the bug), but also reintroduced in old idea we explored in another session that only fuses groups when all the pairwise matches being brought in turn out to be redundant. That isn't the idea behind group fusion; we want the non-redudant pairwise matches to come in, and this is required to make `SectionedCodeTest.reproduceCrossedOverMatches` pass.

2. Although the alignment of blocks has changed in the benchmark merge, that shouldn't in itself be a problem, because the block-level merge only serves to guide the computation of the section-level approximate LCS - it is during the following section-level merge that the filler block gets its chance to be picked up as a left-edit. What should have happened is that the aligned blocks should have resulted in common contributions for the all-sides match, and common to base and right contributions for the pairwise match. In the past, there was some demotion logic for this that exploded aligned blocks into a mixture of common, pairwise common and difference contrinutions, but I think that was possibly simplified into just common and difference contributions. If so, this needs to be finssed to recognise the common pairwise contributions again, and that should open the door for the section-level merge to act correctly.

sageserpent-open:
I’ve dug deeper into the later stages of computing the section-level LCS and merge; it all seems technically correct.

It seems that fusing groups leads to extra alignments being discovered, and for the benchmark merge this leads to an awkward merge outcome.

However, I’m honestly not sure if this is wrong. Perhaps this has just exposed weaknesses in further downstream processing, such as splice merging?

Closing this as inconclusive.

PR #426, opened 2026-09-02
Vet move destinations using parallel matches groups

sageserpent-open:
Updated MoveDestinationsReport.evaluateSpeculativeSourcesAndDestinations to accept a vetting predicate for filtering matches when combining speculative move sources and destinations. In SectionedCodeExtension.scala, constructed the predicate checking whether a match is the first or last element in any parallel matches group. Added unit tests in MoveDestinationsReportTest.

---
*PR created automatically by Jules for task [5193386868571153549](https://jules.google.com/task/5193386868571153549) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
@@ -1184,19 +1184,27 @@ object SectionedCodeExtension extends StrictLogging:
 
       def parallelGroupSectionsFor(
           anchor: Section[Element],
+          blocksFor: Section[Element] => IndexedSeq[Block[Element]],
           sideExtractor: Match[Section[Element]] => Option[Section[Element]]
       ): Set[Section[Element]] =
         val matches: Set[Match[Section[Element]]] =
           sectionedCode.matchesFor(anchor).toSet
-        matches.flatMap { (aMatch: Match[Section[Element]]) =>
-          sectionedCode.parallelMatchesGroupIdsByMatch.get(aMatch).toSeq.flatMap { groupId =>
-            sectionedCode.groupsOfParallelMatches.get(groupId) match
-              case Some(gMatches) =>
-                gMatches.toSeq.flatMap(sideExtractor)
-              case None =>
-                Seq.empty
-          }
+        val groupIds: Set[ParallelMatchesGroupId] = matches.flatMap { aMatch =>
+          sectionedCode.parallelMatchesGroupIdsByMatch.get(aMatch)
         }
+        val blocksOnSide = blocksFor(anchor)
+        val sectionsFromBlocks = blocksOnSide.collect {
+          case block if block.parallelMatchesGroupIds.exists(groupIds.contains) =>
+            block.sectionsCoveredByGroup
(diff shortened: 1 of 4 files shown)


src/test/scala/com/sageserpent/kineticmerge/core/BlockDuplicationAndCondensationTests.scala
```diff
@@ -684,16 +684,16 @@ class BlockDuplicationAndCondensationTests:
       println(s"Right contributions: ${pprintCustomised(rightContributions)}")
 
       println(
-        s"Groups of parallel matches: ${pprintCustomised(sectionedCode.groupsOfParallelMatches)}"
+        s"Groups of parallel matches: ${pprintCustomised(sectionedCode.groupsOfParallelMatches).plainText}"
```
sageserpent-open: Do we need the conversions to plain text now? Was this just for debugging purposes?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1154,18 +1182,38 @@ object SectionedCodeExtension extends StrictLogging:
...
+        val baseGroupSections =
+          parallelGroupSectionsFor(sourceAnchor, _.baseContribution)
+
         def selection(
             candidates: IndexedSeqView[Section[Element]]
         ): IndexedSeq[Section[Element]] =
           candidates
             .takeWhile(candidate =>
-              !basePreservations.contains(candidate) && !sourceAnchors
+              (!basePreservations.contains(candidate) || baseGroupSections
```
sageserpent-open: Regarding this modified clause for the source anchor and the other similar modified clauses for the opposite side anchor and destination anchor, I think the intent here is to distinguish between splices that grow out from the start or end of the implied block, which are terminated by a preservation (or another anchor) and splices that grow into the block, which simply hoover up the block’s sections (again these are terminated by another anchor).

Do I understand correctly?

Regardless, it is worth explaining how the logic works here: could we use either a comment for the entire clause, or two comments for the two subclauses disjoined together, or just introduce some explaining variable bindings?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -1154,18 +1182,38 @@ object SectionedCodeExtension extends StrictLogging:
         selection(file.sections.view.drop(1 + indexOfSection))
       end succeedingAnchoredContentUsingSelection
 
+      def parallelGroupSectionsFor(
```
sageserpent-open, line 1185: Is this enough in all situations? If I understand the use of these sections correctly, then they identify the matched sections in the block associated with the group on the given side.

Should we use the block’s content itself, including any unmatched filler sections in the block? This would be only for a block that has the group id, obviously.

sageserpent-open: I think we can remove `sectionsFromMatches` completely from this function now, as it should be a subset of what is covered by the group’s matches…

sageserpent-open:
Nice work, looks good to me; merging it.

PR #427, opened 2026-09-03
Add --fuse-more-parallel-matches-groups feature toggle

sageserpent-open:
Added a command-line feature toggle --fuse-more-parallel-matches-groups (off by default) that controls whether further parallel matches group fusion takes place in MatchesAndTheirSections.withoutRedundantPairwiseMatches. Redundant pairwise matches are removed regardless of the setting.

---
*PR created automatically by Jules for task [4419406878942423418](https://jules.google.com/task/4419406878942423418) started by @sageserpent-open*

### DIFF
--- a/src/main/scala/com/sageserpent/kineticmerge/Main.scala
+++ b/src/main/scala/com/sageserpent/kineticmerge/Main.scala
@@ -111,6 +111,13 @@ object Main extends StrictLogging:
           .text(
             "Do not commit a successful merge - leave merged changes staged in the index for review. Off by default."
           ),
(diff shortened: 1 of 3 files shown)


sageserpent-open:
I had forgotten that there was a test that failed when additional fusion is not performed.

So it seems that toggling this off is not a good idea after all: while the benchmark merge results seem superficially better when extra fusion doesn’t take place, I’m not keen on allowing users to run the code in a manner that would break tests.

I’ll consider this as a lesson learned; let’s stand down and not pursue this any further.

PR #429, opened 2026-09-05
Add reproduction test for benchmark merge analysis

sageserpent-open:
Adds a minimal reproduction test case in SectionedCodeExtensionTest.scala capturing the multi-occurrence pattern matching stanza behavior in Main.scala during benchmark merges.

---
*PR created automatically by Jules for task [2799586387000257366](https://jules.google.com/task/2799586387000257366) started by @sageserpent-open*


sageserpent-open:
I can't see a concrete reproduction of the original poor merge see in the manual benchmark, so I'm going to try to do this myself by hand.

src/main/scala/com/sageserpent/kineticmerge/core/MatchAnalysis.scala
```diff
@@ -2688,6 +2688,47 @@ object MatchAnalysis extends StrictLogging:
...
+          def checkSpanForSide(
+              sectionExtractor: Match[Section[Element]] => Option[
+                Section[Element]
+              ],
+              sources: Sources[Path, Element],
+              sectionsForSide: collection.Set[Section[Element]]
+          ): Unit =
+            val sectionsOnSide = parallelMatches.toSeq.flatMap(sectionExtractor)
+            if sectionsOnSide.nonEmpty then
+              val filesByPath = sources.filesByPathUtilising(mandatorySections = sectionsForSide.toSet)
```
sageserpent-open: We have to be careful here - if there is a truly alien section, then it may not be recognised because `filesByPathUtilising` will just arbitrarily fill missing sections; such fillers may nest or overlap any alien matches.

sageserpent-open: We could use similar code to what's used in block synthesis here, perhaps? Ideally I'd like the check to stay in the post-reconcilation invariant, so perhaps some shared code could be hoisted?

Alternatively, we could move this check into block synthesis and so do it for each side in isolation. I'm loath to go down this latter route, but if we have to, we can...

sageserpent-open: Just to be clear, the mandatory sections come from: 

```scala
val sectionsOnSide = parallelMatches.toSeq.flatMap(sectionExtractor)
```

That is only the matches from the parallel matches group being vetted. So if there are alien matched sections, these can 'hide' inside the filler sections produced by `filesByPathUtilising`, because these alien matches may possibly group together and / or have gaps between them and the matches from the vetted group.

sageserpent-open: OK, I get it now - I hadn't seen 6d947efefeb6dd34364161ea9ec0f0593b84497b.

Nevertheless, this all feels a bit awkward - while we use valid interfaces to check the invariant, it feels as if `MatchesAndTheirSections` is trying emulate its higher-level client class `SectionedCode` when it calls `filesByPathUtilising`.

Here's an idea - can we use the likes of `baseSectionsByPath`, `leftSectionsByPath`, `rightSectionsByPath`, `sectionsAndTheirMatches` and `parallelMatchesGroupIdsByMatch` to iterate through each path on each side and confirm that the sections on each path that are matched proceed in a sequence where group id affiliation doesn't have 'gaps' where an id drops out, only to reappear later?

In effect we invert the check to work from the POV of the sections on a path, looking for inconsistent group affiliations, rather than from the POV of groups, looking for inconsistent sections.

That would keep the invariant check more localised, albeit at the cost of working with the more concrete representation in `MatchesAndTheirSections`.

PR #438, opened 2026-09-13
Port Main.scala to Kyo effect system

sageserpent-open:
Ports Main.scala from Cats effect monad stack (IO, EitherT, WriterT) to the Kyo effect library (kyo.IO, Abort, Emit). Collection traversals have been updated to use Kyo extension helpers, and temporary bridge code has been removed. All tests in MainTest pass cleanly.

---
*PR created automatically by Jules for task [11140145498289177921](https://jules.google.com/task/11140145498289177921) started by @sageserpent-open*


src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -367,47 +364,55 @@ object Main extends StrictLogging:
...
-        WriterT.value
+    import AllowUnsafe.embrace.danger
+
+    val (log, exitCode) = (kyo.IO.Unsafe.evalOrThrow(
+      Emit.run(
+        Abort.run[String](workflow)
       )
-      .run
-      .unsafeRunSync()
+    ): @unchecked) match
```
sageserpent-open, line 373: The logic here is an improvement on what was before: I can actually understand it!

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -275,12 +272,12 @@ object Main extends StrictLogging:
     )
 
     val workflow = for
-      _ <- IO {
+      _ <- kyo.IO {
```
sageserpent-open, line 275: Do we have to have the `kyo` package prefix? Is the Cats `IO` still leaking in?

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -367,47 +364,55 @@ object Main extends StrictLogging:
...
-        .leftMap(exception =>
-          // TODO: something pure, functional and wholesome that could be seen
-          // at high church...
-          logger.error(exception.getMessage)
-          exception.printStackTrace()
-          errorMessage.taggedWith[Tags.ErrorMessage]
-        )
+  extension [Payload](fallible: Payload < kyo.IO)
+    private def labelExceptionWith(errorMessage: String): KyoWorkflow[Payload] =
+      val handled: Result[Throwable, Payload] < kyo.IO = Abort.run[Throwable](fallible)
```
sageserpent-open, line 397: This doesn't feel right - it evaluates one of the effects, only to inspect its payload and then wrap it up in an effect again. Is there a more idiomatic way of achieving this? This is essentially burying an imperative evaluation in the middle of pure monadic code...

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -367,47 +364,55 @@ object Main extends StrictLogging:
...
+      }
   end extension
 
-  extension [Payload](workflow: Workflow[Payload])
-    private def logOperation(message: String): Workflow[Payload] =
-      workflow.semiflatTap(_ => WriterT.tell(List(Right(message))))
+  extension [Payload](workflow: KyoWorkflow[Payload])
+    private def logOperation(message: String): KyoWorkflow[Payload] =
+      workflow.flatMap { payload =>
+        Emit.value[ErrorOrOperationMessage](Right(message)).map(_ => payload)
```
sageserpent-open, line 416: Does Kyo have the equivalent of Cat's `as` helper to make the intent more obvious here?

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -33,10 +30,10 @@ object Main extends StrictLogging:
...
   private type ErrorOrOperationMessage =
     Either[String @@ Tags.ErrorMessage, String]
-  private type WorkflowLog                = List[ErrorOrOperationMessage]
-  private type WorkflowLogWriter[Payload] = WriterT[IO, WorkflowLog, Payload]
-  private type Workflow[Payload]          =
-    EitherT[WorkflowLogWriter, String @@ Tags.ErrorMessage, Payload]
+
+  // Kyo-based workflow effect type:
+  type KyoWorkflow[Payload] =
+    Payload < (kyo.IO & Abort[String] & Emit[ErrorOrOperationMessage])
```
sageserpent-open, line 36: Shouldn't this be `Abort[String @@ Tags.ErrorMessage]` for parity?

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -426,16 +431,14 @@ object Main extends StrictLogging:
     commandLineArguments = commandLineArguments*
   )
 
-  private def left[Payload](errorMessage: String): Workflow[Payload] =
-    EitherT.leftT[WorkflowLogWriter, Payload](
-      errorMessage.taggedWith[Tags.ErrorMessage]
-    )
+  private def left[Payload](errorMessage: String): KyoWorkflow[Payload] =
+    Abort.fail[String](errorMessage)
```
sageserpent-open, line 437: This works, but I'd prefer (as per an earlier review comment) to tag `errorMessage` here rather than later on as errors are extracted.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -673,42 +676,35 @@ object Main extends StrictLogging:
...
+    ): KyoWorkflow[(Path, Change)] =
+      line.split(whitespaceRun) match
+        case Array("M", changedFile) =>
+          val path = workingDirectory / RelPath(changedFile)
+          for
+            modeAndBlob <- blobFor(commitIdOrBranchName)(path)
+            (mode, blobId) = modeAndBlob
+            contentOpt <-
+              if binaryContentInvolvedFor(path) then right(Option.empty[String @@ Tags.Content])
+              else contentFor(commitIdOrBranchName, path)(blobId).map(c => Some(c))
```
sageserpent-open: Using `Some.apply` would be pithier...

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -673,42 +676,35 @@ object Main extends StrictLogging:
...
+          yield path -> Change.Addition(mode, blobId, contentOpt)
+
+        case Array("D", deletedFile) =>
+          val path = workingDirectory / RelPath(deletedFile)
+          right(path -> Change.Deletion(binaryContentInvolvedFor(path)))
+
+        case _ =>
+          left(s"Unexpected error - can't parse changes reported by Git ${underline(line)}.")
```
sageserpent-open, line 712: Is the equivalent to the old code - that translated an intercepted exception, whereas this is just adding another pattern matching arm - one what I don't think will be executed.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -717,283 +713,241 @@ object Main extends StrictLogging:
...
+            mergedFileMode <-
+              if bestAncestorCommitIdMode == ourModification.mode then
+                right(theirModification.mode)
+              else if bestAncestorCommitIdMode == theirModification.mode then
+                right(ourModification.mode)
+              else if ourModification.mode == theirModification.mode then
+                right(ourModification.mode)
+              else
+                left(
+                  s"Conflicting file modes for file ${underline(path)}; on best ancestor commit ${underline(bestAncestorCommitIdMode)}, on our branch head ${underline(ourModification.mode)} and on their branch head ${underline(theirBranchHead)}."
```
sageserpent-open: No, this is wrong - it is the their file's *mode* that is being shown, not the name of their branch.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -1009,11 +963,11 @@ object Main extends StrictLogging:
         commitIdOrBranchName: String @@ Tags.CommitOrBranchName
     )(
         path: Path
-    ): Workflow[
+    ): KyoWorkflow[
       (String @@ Tags.Mode, String @@ Tags.BlobId)
     ] =
       for
-        Array(mode, entryType, entryId, _) <- IO {
+        parsed <- kyo.IO {
```
sageserpent-open: The direct pattern unpacking was a lot easier to read.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -1795,29 +1755,32 @@ object Main extends StrictLogging:
...
-            )
-          }
-          .leftMap(_.toString.taggedWith[Tags.ErrorMessage])
+        sectionedCode: SectionedCode[Path, Token] <- SectionedCode.of(
+          baseSources,
+          leftSources,
+          rightSources
+        )(
+          configuration.copy(label = "Match analysis")
+        ).fold(
```
sageserpent-open: Would `Abort.get` work better here?

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -1795,29 +1755,32 @@ object Main extends StrictLogging:
...
+        )
 
         (mergeResultsByPath, moveDestinationsReport) =
           given ProgressRecording = configuration.progressRecording
 
           sectionedCode.merge
 
-        _ <- moveDestinationsReport.summarizeInText.foldLeft(right(()))(
-          _ `logOperation` _
+        _ <- Kyo.foreach(moveDestinationsReport.summarizeInText)(msg =>
```
sageserpent-open, line 1777: Won't the old approach work here, given that `logOperation` has been cut over? I'd rather stick with using that helper function as it makes the intent clearer.

sageserpent-open (commented):
This looks promising, but further work needs to be done. I'm also unsure as to whether the handling of what used to be sequenced `Option` values is easier to follow now.

A question I have about Kyo is, will the Kyo `IO` monad absorb exceptions in the same way as the Cats one does? If not, any exception will bomb out of the execution straightaway.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: I haven't seen this go back to an unprefixed `IO` yet, at least not in commit 6ce818dd984d13d7b9e5cf6b80b492bd51cadc7b.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -367,47 +364,55 @@ object Main extends StrictLogging:
...
   private def underline(anything: Any): Str =
     fansi.Underlined.On(anything.toString)
 
-  extension [Payload](fallible: IO[Payload])
-    private def labelExceptionWith(errorMessage: String): Workflow[Payload] =
-      EitherT
-        .liftAttemptK[WorkflowLogWriter, Throwable]
-        .apply(WriterT.liftF(fallible))
-        .leftMap(exception =>
-          // TODO: something pure, functional and wholesome that could be seen
```
sageserpent-open, line 400: This comment has been deleted, but it still holds for the new code, because of the call to `logger.error` / `exception.printStackTrace`.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: Hang on - my point was that the code *is* running imperative actions (via `Abort.run`) in the middle of what looks like pure monadic code. To be clear, it's not the call to `logger.error` that bothers me here - although that has its own problems, as noted by another other review comment wrt the old code.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: The use of `Emit.value` is fine, but I was hoping to hide the cryptic `.map(_ => payload)`  in som equivalent to Cat's `as` helper.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: I can't see any further changes as of commit 6ce818dd984d13d7b9e5cf6b80b492bd51cadc7b.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: Again, I can' t see any further changes as of commit 6ce818dd984d13d7b9e5cf6b80b492bd51cadc7b.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: Again, I can't see any further changes as of commit 6ce818dd984d13d7b9e5cf6b80b492bd51cadc7b.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: This hasn't been updated, though. I was expecting to see `logOperation` back in use here.

sageserpent-open:
There are still unresolved issues with this PR, although they are claimed to be fixed.

I'm not sure if this is a problem with staging Jules' local changes in the session.

For now, I'm treating this as a useful starting point if the decision is made to cut over to Kyo, but I'm not merging this as-is.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
sageserpent-open: It has switched back to showing the branch name again.

PR #440, opened 2026-09-16
Demote AllSides matches when cross-set crossovers occur

sageserpent-open:
Resolves test failure in SectionedCodeTest.reproduceCrossedOverMatches by demoting AllSides matches when cross-set crossovers occur.

---
*PR created automatically by Jules for task [1268450999418751999](https://jules.google.com/task/1268450999418751999) started by @sageserpent-open*


src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -681,23 +681,78 @@ object SectionedCodeExtension extends StrictLogging:
...
+
+        def tryDemoteToResolve(
+            mToDemote: Match[Section[Element]],
+            mOther: Match[Section[Element]],
+            matches: Seq[Match[Section[Element]]]
+        ): Option[Seq[Match[Section[Element]]]] =
+          mToDemote match
+            case Match.AllSides(b, l, r) =>
+              val demotionCandidates = Seq(
+                (Match.BaseAndLeft(b, l), Option(r)),
```
sageserpent-open: Why are the freed sections wrapped in `Option`?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -681,23 +681,78 @@ object SectionedCodeExtension extends StrictLogging:
...
+              resolveCrossovers(resolution)
+
+        val rawMatches = resolveCrossovers(initialMatches)
+
         rawMatches.combinations(2).foreach {
           case Seq(m1, m2) =>
             assert(
               !matchesCross(m1, m2),
-              s"Post-condition failed: matches cross!\n$m1\n$m2"
+              s"Post-condition failed: matches cross!\n\n"
```
sageserpent-open, line 746: Why not continue to report which matches have crossed over?

sageserpent-open (commented):
Some initial thoughts while I’m trying to fully understand what was done here…

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -681,23 +681,76 @@ object SectionedCodeExtension extends StrictLogging:
...
           )
 
+        def findCrossingPair(matches: Seq[Match[Section[Element]]]): Option[(Match[Section[Element]], Match[Section[Element]])] =
+          matches.combinations(2).collectFirst {
+            case Seq(m1, m2) if matchesCross(m1, m2) => (m1, m2)
+          }
+
+        def tryDemoteToResolve(
+            mToDemote: Match[Section[Element]],
+            mOther: Match[Section[Element]],
```
sageserpent-open: `mOther` isn't used.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -681,23 +681,75 @@ object SectionedCodeExtension extends StrictLogging:
...
+              val demotionCandidates = Seq(
+                (Match.BaseAndLeft(b, l), r),
+                (Match.BaseAndRight(b, r), l),
+                (Match.LeftAndRight(l, r), b)
+              )
+
+              demotionCandidates.iterator.flatMap { case (demoted, freedSection) =>
+                val remaining = matches.filterNot(_ == mToDemote)
+                if !remaining.exists(matchesCross(demoted, _)) then
+                  val replacementOpt = matchSequence.find { candidate =>
```
sageserpent-open: I'm wonder about this `replacementOpt`. It seems that an attempt is made to find another alignment match that holds the section freed from the demoted match, one that doesn't cross with the resolved matches.

The thing is, if such a match exists (and it could do) in `matchSequence`, then won't it also be considered by the resolution process in turn?

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -681,23 +681,64 @@ object SectionedCodeExtension extends StrictLogging:
...
+        ): Option[Seq[Match[Section[Element]]]] =
+          mToDemote match
+            case Match.AllSides(b, l, r) =>
+              val demotionCandidates = Seq(
+                (Match.BaseAndLeft(b, l), r),
+                (Match.BaseAndRight(b, r), l),
+                (Match.LeftAndRight(l, r), b)
+              )
+
+              demotionCandidates.iterator.flatMap { case (demoted, freedSection) =>
```
sageserpent-open: Having simplified this function, we may as well dispense with `freedSection`...

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
sageserpent-open: Nice touch using `find`.

I suggest hoisting `val remaining = matches.filterNot(_ == mToDemote)` out of the predicate so it can be used in the following `.map` call.

sageserpent-open:
This looks good to me: I've run `com.sageserpent.kineticmerge.core.SectionedCodeTest#reproduceCrossedOverMatches` with the debugger and can confirm that demotion does actually occur.

However, I've reinstated an earlier approach to resolving crossed-over alignment matches that was introduced as part of a larger PR and subsequently reverted (because of performance issues resulting from wider changes in that PR), this is brought back in commit 3804ab505a75ac273fec1b39c6b5375a01b36466.

I'm going to think about the trade-off between both approaches, and whether they can be combined together.

Also of interest is whether the cross-over detection can be made more efficient by taking sorting the alignment matches (which are all on the same path on both sides) from the POV of each side, and then merging the sorted sequences in turn to detect cross-overs as out of sequence items (or alternatively, using LCS on the aligned matches sequences).

Let's put this on hold for now and cogitate...

sageserpent-open:
I've merged the other approach to removing crossed-over alignment matches in commit 320e57819aebf0e04edb6a03f029dd6eebfbe791.

Despite increasing the time budgets for `SectionedCodeTest.sourcesCanBeReconstructedFromTheAnalysis` and `SectionedCodeTest.matchingSectionsAreFound` to 20 minutes, I have yet to see even the demotion logic actually be executed, let alone the fallback logic that now chooses the largest match.

It seems that alignment match crossovers are now extremely rare, which is a good state of affairs.

Consequently I'm going to take this work as being extremely defensive coding (albeit with a single test that exercises the demotion logic path only, but not the fallback path).

Once the CI build goes green I shall merge this. Optimising the crossover detection can be followed up later.

PR #443, opened 2026-09-17
Use LCS to resolve crossovers in alignment matches

sageserpent-open:
Replaced the tail-recursive `resolveCrossovers` function in `SectionedCodeExtension.scala` with a turn-by-turn alignment approach using Kinetic Merge's existing `LongestCommonSubsequence.of` algorithm operating on each side's sorted match sequences. AllSides matches are demoted to pairwise matches or eliminated based on LCS contribution types, and all existing tests pass cleanly.

---
*PR created automatically by Jules for task [16842061981776765962](https://jules.google.com/task/16842061981776765962) started by @sageserpent-open*


src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -695,49 +695,78 @@ object SectionedCodeExtension extends StrictLogging:
               representativeMatchesFrom(matchesSharingASectionOnAtLeastOneSide)
             )
 
-        def findCrossingPair(
-            matches: Seq[Match[Section[Element]]]
-        ): Option[(Match[Section[Element]], Match[Section[Element]])] =
-          matches.combinations(2).collectFirst {
-            case Seq(m1, m2) if matchesCross(m1, m2) => (m1, m2)
-          }
+        given Sized[Match[Section[Element]]] = _.size
```
sageserpent-open: These given declarations seem correct, but I had a doubletake when I saw the `Order` given before the match sorting - why did we need the `Order` when there is an explicit `.sortBy`?

The answer is that it is needed by the LCS invocation. I suggest moving the two given definitions closer to the call site to make the connection more obvious.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
```diff
@@ -695,49 +695,78 @@ object SectionedCodeExtension extends StrictLogging:
...
+            aMatch.leftContribution.map(_.startOffset),
+            aMatch.rightContribution.map(_.startOffset)
+          )
+        )
+
+        val baseMatches = initialMatches
+          .flatMap(m => m.baseContribution.map(_ => m))
+          .sortBy(m =>
+            (
+              m.baseContribution.get.startOffset,
```
sageserpent-open: Ah, I see there is a tiebreaker ordering going on here - make sense in case start offsets collide.

Nevertheless, could the permutation of the tiebreaking orderings for the three sides cause spurious misalignments if the start offsets collide on more than one side?

I'm unsure about this, because a) the side whose perspective is being taken has to come first - that's what we're really after. Is it enough that the remaining parts of the tiebreak are always a subsequence of base, then left, then right? Hmm...

sageserpent-open (commented):
I like this, but I have a minor nitpick plus a question about the correctness.

src/main/scala/com/sageserpent/kineticmerge/core/SectionedCodeExtension.scala
sageserpent-open: I was thinking of a similar situation to what goes in when the parallel matches groups are checked for order consistency in the reconcilation post-condition (in `MatchesAndTheirSections`) - their matches are presorted first in case of collisions on one side, then sorted wrt to the 'perspective side'. Maybe we should do the same thing here?

My concern was that if multiple matches shared the same start offset on two sides, they might be ordered differently due to the tiebreaking permutations, and that could show up as a synthetic misalignment going into the LCS inputs.

sageserpent-open:
Looks good to me, I’ll merge this.

PR #444, opened 2026-09-17
Bypass alignment match LCS when no crossovers are present

sageserpent-open:
Introduces an O(N) crossover check before running match alignment LCS, bypassing expensive 3D dynamic programming grid calculations when no crossovers exist.

---
*PR created automatically by Jules for task [1829506417482417279](https://jules.google.com/task/1829506417482417279) started by @sageserpent-open*


src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
```diff
@@ -127,6 +127,23 @@ case class LongestCommonSubsequence[Element] private (
         )
       )
 
+  def concat(
```
sageserpent-open, line 130: An old friend returns - I remember something similar that was removed back in afc1136544e9c7336576372c84a3dada17c691f0. You can't keep a good method down!

src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
```diff
@@ -358,98 +375,434 @@ object LongestCommonSubsequence:
...
+        end if
       end while
 
-      index
-    end commonAffixLength
+      result
+    end commonPrefixContributions
 
-    val minimumSizeAcrossSides = base.size min left.size min right.size
+    def commonSuffixContributions(
```
sageserpent-open, line 650: This helper has more or less the same logic as in `commonPrefixContributions`. Can we do something like what used to be done with the old `commonAffixLength` and work on `IndexedSeqView[Element]`, paasing in reversed sequences for the suffix calculation?

Obviously, we'll have to reverse the resulting `LongestCommonSubsequence[Element]`, but it would be easier to review at least.

src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
```diff
@@ -358,98 +375,434 @@ object LongestCommonSubsequence:
...
+            end if
+          else if equality.eqv(l, r) then
+            if !baseSet.contains(l) then
+              result = result.addCommonLeftAndRight(l, r)(sized.sizeOf)
+              leftIdx += 1
+              rightIdx += 1
+            else keepGoing = false
+            end if
+          else
+            if !leftSet.contains(b) && !rightSet.contains(b) then
```
sageserpent-open: I understand the logic here - this is the same idea that was employed down in `assumingTrimmedInputs`, outlined in the comment on line 811. Here, we use this reasoning to extend the prefix through mismatches where an element can't possibly be matched on any other side.

The other logic isn't made redundant by this, because here we are still in the prefix (or suffix for the analogous function further down); the other logic works on the 'core' piece of the LCS that's left over.

Nevertheless, wouldn't it be better to hoist the related logic from line 811 right up into the outermost layer of optimisation (I would recommend writing a new optimisation layer function for this) so that it is the very first optimisation performed on the initial LCS inputs. That way we can remove the conceptual duplication at this level and down in `assumingTrimmedInputs`.

sageserpent-open:
This is good enough to merge; both the session benchmarking and local runs on my own machine indicate the absence of any swathe-level LCS calculations for aligned matches and a considerable consequent speedup wrt to what's on `main` at the moment.

Further comparative benchmarking also reveals no performance regression wrt commit a8314d39536680e612f3a71aba624969dd4cc1a2, which was just before the aligned-matches LCS was introduced as a way of detecting and demoting crossovers.

I'm still not convinced that we can't finesse the fusion of `commonPrefixContributions` and `commonSuffixContributions` as per the approach used in the code prior to this PR, but that's not worth blocking this PR any further; I'm merging it.

src/main/scala/com/sageserpent/kineticmerge/core/LongestCommonSubsequence.scala
sageserpent-open: I can pick this up one the PR is merged.

PR #445, opened 2026-09-18
Cut over Main.scala Workflow functions to direct style using dotty-cps-async

sageserpent-open:
Replaced for-comprehensions across all Workflow monadic functions in Main.scala with direct-style reify/reflect constructs powered by dotty-cps-async. Added a CpsMonad instance for the Workflow monad transformer stack and added the dotty-cps-async dependency to build.sbt.

---
*PR created automatically by Jules for task [12720160508459589193](https://jules.google.com/task/12720160508459589193) started by @sageserpent-open*


src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -959,31 +990,34 @@ object Main extends StrictLogging:
...
                       s"but our branch thinks the original is ${description(binaryContentDeletedOnLeft)} " +
                       s"and their branch thinks the original is ${description(binaryContentDeletedOnRight)}."
                   )
-                else right(())
-              bestAncestorCommitIdContent <- Option
-                .unless(binaryContentDeletedOnLeft)(
-                  contentFor(bestAncestorCommitId, path)(
-                    bestAncestorCommitIdBlobId
-                  )
                 )
```
sageserpent-open: Dropping the else branch is an improvement!

sageserpent-open:
I've reformatted `Main.scala` to see what impact the changes made on indentation level - as it turns out, not much! Pushed commit 2092e0fe3a61e8154aaee96563ef556e922be73a to pick this up in the PR.

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -435,17 +467,20 @@ object Main extends StrictLogging:
...
       content: String @@ Tags.Content
   ): Workflow[Path] =
-    for temporaryFile <- IO {
-        os.temp(
-          contents = content,
-          prefix = "kinetic-merge-",
-          suffix = ".base",
-          deleteOnExit = true
+    reify {
+      reflect(
```
sageserpent-open: It seems that the original for-comprehension was redundant anyway - I suppose it just called out the exception labelling. We may as well dispense with the `reify`/`reflect` nesting here and just return the labelled `IO` block directly....

src/main/scala/com/sageserpent/kineticmerge/Main.scala
```diff
@@ -1481,20 +1572,25 @@ object Main extends StrictLogging:
           conflictingDeletedPathsByLeftRenamePath.toSeq
             .foldM(this) {
               case (partialResult, (leftRenamedPath, conflictingDeletedPath)) =>
-                for
-                  _              <- recordDeletionInIndex(leftRenamedPath)
-                  (mode, blobId) <- blobFor(ourBranchHead)(
-                    leftRenamedPath
+                reify {
+                  reflect(recordDeletionInIndex(leftRenamedPath))
```
sageserpent-open: This is easier on the eye, good.

sageserpent-open (commented):
Looks quite promising so far. I have one minor nitpick so far, but that's it.

sageserpent-open:
Once we have that nitpick sorted out, it would be nice as a stretch goal to try out https://dotty-cps-async.github.io/dotty-cps-async/BasicUsage.html#direct-context-encoding-experimental to cut down on the noise from the `reify` / `reflect` invocations.

sageserpent-open:
I think we can tidy up the use of the `reflect` calls a little to make things easier to read.

For example:

```scala
    val workflow = reify {
      reflect(IO {
        os.proc("git", "--version").call(workingDirectory)
      }.labelExceptionWith(errorMessage = "Git is not available."))

      val topLevel = reflect(
        IO {
          os.proc("git", "rev-parse", "--show-toplevel")
            .call(workingDirectory)
            .out
            .text()
            .strip()
        }.labelExceptionWith(errorMessage =
          "The current working directory is not part of a Git working tree."
        )
      )

      ...
```

Can be rewritten as:

```scala
      IO {
        os.proc("git", "--version").call(workingDirectory)
      }.labelExceptionWith(errorMessage = "Git is not available.").reflect

      val topLevel =
        IO {
          os.proc("git", "rev-parse", "--show-toplevel")
            .call(workingDirectory)
            .out
            .text()
            .strip()
        }.labelExceptionWith(errorMessage =
          "The current working directory is not part of a Git working tree."
        ).reflect
```

That is slightly less eye-watering.

Can we cut these `reflect` calls over to use the explicit extension call syntax?

